PR #347, opened 2026-05-25
docs: remove unused grammar idents from enums macro

### LINKED ISSUE #344: enums: Cpp variant declares tree_sitter_cpp but resolves to tree_sitter_mozcpp
> **Status:** Fixed on `fix/issue-344` by commit
> 4595f2aeb63bf396572747c20c6d18b95c89817e (Option A). Awaiting
> integration; do not close until the parent batch lands.

## Summary

`enums/src/languages.rs` lists `(Cpp, tree_sitter_cpp)` as a
`mk_langs!` entry, but the macro discards the second tuple element
and `get_language` actually returns `tree_sitter_mozcpp::LANGUAGE`.
The token-table emitted for `Cpp` is therefore derived from
`tree_sitter_mozcpp`, not `tree_sitter_cpp`, despite what the source
declares.

## Location

- `enums/src/languages.rs:12` — declaration
- `enums/src/macros.rs:18-46` — macro that discards `$name`
- `enums/src/macros.rs:37` — actual `Lang::Cpp => tree_sitter_mozcpp::LANGUAGE.into()`

## Evidence

`enums/src/languages.rs:5-12`:

```rust
mk_langs!(
    // 1) Name for enum
    // 2) tree-sitter function to call to get a Language
    (Kotlin, tree_sitter_kotlin_ng),
    ...
    (Cpp, tree_sitter_cpp),
```

`enums/src/macros.rs:18-44`:

```rust
macro_rules! mk_get_language {
    ( $( ($camel:ident, $name:ident) ),* ) => {
        pub fn get_language(lang: &Lang) -> Language {
            match lang {
                Lang::Kotlin => tree_sitter_kotlin_ng::LANGUAGE.into(),
                ...
                Lang::Cpp => tree_sitter_mozcpp::LANGUAGE.into(),
                ...
```

The macro arm captures `$name:ident` but the body never expands
`$name` — every grammar path is hardcoded. The second tuple element
in `languages.rs` is therefore decorative; the only place it
matters is as a per-call comment for a human reader, and for `Cpp`
that comment is wrong.

Verified via `cargo tree`: `tree-sitter-cpp v0.23.4` is pulled in
only as a transitive dependency of `bca-tree-sitter-mozcpp`, not
because `enums/Cargo.toml` depends on it directly. The crate has no
direct `tree-sitter-cpp` dependency.

## Expected Behavior

Either:

1. The macro actually uses `$name` (so adding a new language is
   a one-line edit to `languages.rs` instead of two), OR
2. The second tuple element accurately documents which crate
   `get_language` resolves the variant against.

## Actual Behavior

The declaration says `tree_sitter_cpp` but the resolved grammar is
`tree_sitter_mozcpp`. The macro signature implies the relationship
is meaningful when it is not.

## Impact

- Contributor confusion: a reader of `languages.rs` would
  reasonably expect `Cpp` token tables to be generated against the
  vanilla `tree-sitter-cpp` grammar, when in fact they reflect
  `tree-sitter-mozcpp` (which has Mozilla-specific divergences).
- Drift risk: if a future bump to `tree-sitter-cpp` lands in the
  workspace `Cargo.toml`, this stale declaration provides no
  enforcement.

## Resolution (Option A)

Dropped the unused `$name:ident` capture from `mk_get_language!`
and `mk_langs!`. Each entry in `languages.rs` is now a bare
`$camel` ident; non-obvious mappings carry an inline
`// -> <crate>` annotation directly attached to the right Lang
(`#[rustfmt::skip]` on the invocation prevents rustfmt from
collapsing neighbouring lines and reattaching comments to the
wrong variant). No public API change: `Lang::Cpp` and friends
still resolve exactly as before.

Option B (wiring `$name` through so the declaration becomes
load-bearing) was rejected because it would have required
renaming `Lang::Cpp` to `Lang::Mozcpp` or introducing an `as`
alias on the grammar import — both ripple into the published
re-export surface and exceed the scope of the bug.

narutamaaurum:
## Summary
- remove the unused second tuple element from `enums/src/languages.rs`
- keep grammar resolution centralized in `mk_get_language!` instead of implying the list is load-bearing
- add an explicit note that `Cpp` intentionally resolves to `tree_sitter_mozcpp::LANGUAGE`

## Validation
- `git diff --check`
- `cargo test --manifest-path enums/Cargo.toml` *(could not run in this environment: `cargo` is not installed)*

Closes #344

### DIFF
--- a/enums/src/languages.rs
+++ b/enums/src/languages.rs
@@ -1,27 +1,27 @@
 use tree_sitter::Language;
 
 mk_langs!(
-    // 1) Name for enum
-    // 2) tree-sitter function to call to get a Language
-    (Kotlin, tree_sitter_kotlin_ng),
-    (Lua, tree_sitter_lua),
-    (Java, tree_sitter_java),
-    (Go, tree_sitter_go),
-    (Rust, tree_sitter_rust),
-    (Tcl, tree_sitter_tcl),
-    (Cpp, tree_sitter_cpp),
-    (Python, tree_sitter_python),
-    (Tsx, tree_sitter_tsx),
-    (Typescript, tree_sitter_typescript),
-    (Bash, tree_sitter_bash),
-    (Csharp, tree_sitter_c_sharp),
-    (Elixir, tree_sitter_elixir),
-    (Ccomment, tree_sitter_ccomment),
-    (Preproc, tree_sitter_preproc),
-    (Mozjs, tree_sitter_mozjs),
-    (Javascript, tree_sitter_javascript),
-    (Perl, tree_sitter_perl),
-    (Php, tree_sitter_php),
-    (Ruby, tree_sitter_ruby),
-    (Groovy, dekobon_tree_sitter_groovy)
+    // Enum variants only. Grammar resolution lives in mk_get_language!.
+    (Kotlin),
+    (Lua),
+    (Java),
+    (Go),
+    (Rust),
+    (Tcl),
+    // Cpp intentionally resolves to tree_sitter_mozcpp::LANGUAGE.
+    (Cpp),
+    (Python),
+    (Tsx),
+    (Typescript),
+    (Bash),
+    (Csharp),
+    (Elixir),
+    (Ccomment),
+    (Preproc),
+    (Mozjs),
+    (Javascript),
+    (Perl),
+    (Php),
+    (Ruby),
+    (Groovy)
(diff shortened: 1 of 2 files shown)


dekobon:
Already fixed in 3f838f507ef798b4e78ea96a3d9b6f83d15f579f

Issue #736, opened 2026-06-13
track(quality): metric-driven maintainability cleanup plan


dekobon:
> **Status (2026-06-14):** Tier 1 + Tier 2 ✅ **merged** via #745. Tier 3
> **assessed and closed as wontfix** — all three targets are large data/render
> files with no per-function complexity (only the file-volume `loc.sloc` /
> `halstead.effort` metric trips, and they are already baselined). Refactoring
> them purely to shrink a file-size number would be the metric-driven churn the
> anti-gaming policy warns against. See the closing comment for the per-file
> assessment. **Closed.**

---

## Summary

A metric-driven audit of the workspace, using `bca` against itself (the
dogfood gate). This issue collects the highest-value maintainability
improvements the numbers point to, with each candidate triaged into
*genuine refactor* vs *essential-complexity-suppress* so we **fix the
code, not the number** (per the "Responding to bca metric feedback"
policy in `AGENTS.md`).

## How the data was gathered

- `bca report --format markdown --top 25` over the whole repo (honouring
  `bca.toml`: `exclude_tests = true`, `cyclomatic_count_try = false`).
- `bca check` (hard gate) + `bca exemptions` (baseline + suppression audit).
- Snapshot: 135 files, 68,013 SLOC, 3,094 functions, 524 types.
  Rust dominates (129 files / 67k SLOC); Bash (3) and Python (3) are
  clean (`No major quality concerns detected`).

## Caveats — what is NOT debt to chase here

1. **The current live `bca check` regressions are not pre-existing debt.**
   The three violations on a dirty tree —
   `markdown_report.rs::language_display_name` cyclomatic 26 (baseline 25),
   `alterator.rs` wmc 66 (baseline 61), `metrics/nargs.rs` wmc 70
   (baseline 59) — are caused entirely by the **uncommitted Objective-C
   language addition** (`alterator.rs` +31, `nargs.rs` +170, 21 `objc`
   refs). These belong to that PR and just need a baseline refresh
   (`make self-scan-write-baseline-headroom`) in the same change. Listed
   only so they are not double-counted as debt.
2. **Large file SLOC is mostly test-inflated.** Rust co-locates
   `#[cfg(test)]` modules in-file; the `report` SLOC column is the file
   root span and includes them, so headline file sizes (e.g.
   `metrics/abc.rs` 4,313, `getter.rs` 2,649) overstate production size.
   The gate already prunes tests for node-counted metrics; the per-file
[...]

Issue #705, opened 2026-06-10
preproc C/C++: #undef recorded as defined, library eprintln, ambiguous include fan-out


dekobon:
Found during a systematic per-file source-review sweep. `src/preproc.rs` (C/C++ #define/#include resolution). Cycle handling and path discipline are sound; these are the substantive issues.

- **`#undef` is recorded as a DEFINED macro** — `:359-372`. The arm `Preproc::Define | Preproc::Undef =>` unconditionally inserts the identifier into `file_result.macros`, so `#undef FOO` *adds* FOO instead of removing it; `get_macros` then propagates it to includers → wrong macro set feeding the C/C++ `c_macro::replace` expansion → potentially wrong C metrics. Zero `#undef` test coverage. Confirm intent vs the consumer; if "undef means no longer defined," the two directives must not share an arm.
- **Library writes `eprintln!` diagnostics unconditionally** — `:128,196-199,207-214,253-257,263-266`. The line-127 `// TODO: add an option to display warning` confirms it; an embedder (`bca-web`) can't suppress/capture self-inclusion/cycle/skip warnings. Route through a log/tracing facade or return collected diagnostics.
- **Ambiguous `#include` resolves to MULTIPLE candidates, each added as a graph edge** — `:120-131` (+ tools.rs `min_distance_candidates`). A basename like `config.h` in several dirs gains edges to all tied candidates → macros leak from unrelated files via `get_macros`; candidate order also depends on `all_files` Vec ordering. Add a deterministic tie-break + document the best-effort caveat.
- No project-root containment on include resolution (path traversal, low severity — resolution is a lookup into pre-collected `all_files`, not a filesystem open) — `:121`.

---

### Locked decisions (design audit, 2026-06-10)

- **`#undef` un-defines** — split the shared `Define | Undef` arm
  (`preproc.rs:359-372`): `#undef FOO` removes FOO from `file_result.macros`
  (and must override an includer's define in the propagated set, so verify how
  `c_macro::replace` consumes the set and handle define-then-undef ordering
  within a file). Add `#undef` test coverage (currently zero).
- **Diagnostics: return collected diagnostics to the caller** — replace the
  unconditional `eprintln!`s (`:128,196-199,207-214,253-257,263-266`; the
  `:127` TODO) with warnings threaded out as returned data. CLI prints them,
  bca-web captures/suppresses, tests assert on them. Accept the preproc
  return-signature change threaded through callers.
[...]

PR #745, opened 2026-06-14
refactor(metrics): reduce complexity hotspots from self-scan

dekobon:
## Summary

Behavior-preserving refactors of the per-function complexity hotspots
surfaced by the project's own `bca` self-scan (see #736 for the audit and
plan). Each oversized function is split into named, single-purpose helpers
at a boundary a reader would draw — the metric drops because the code is
genuinely simpler, not because a `foo_part2` helper moved the count.

Implements **Tier 1 + Tier 2** of #736. Tier 3 (`web/negotiate.rs`, the
large-but-cohesive report builders) is intentionally deferred — it was the
"leave unless a natural boundary emerges" tier.

## Changes & measured impact

All numbers measured with the repo's own `bca` (honoring `bca.toml`).

| Target | Metric | Before → After | How |
| --- | --- | --- | --- |
| `preproc.rs::preprocess_with_parser` | cognitive | **33 → ~10** | split into `push_children` / `classify_preproc_node` / `apply_macro_events` |
| `preproc.rs::record_indirect_includes` | cognitive | **25 → ~10** | extracted `accumulate_reachable_includes` |
| `cli/lib.rs::expand_seed_paths` | cognitive | **24 → 11** | extracted `walk_directory_seed` (file-seed vs dir-walk split) |
| `cli/commands.rs::run_command_report` | ABC | **46 → 41** | collapsed redundant 4-arm `(format, vcs)` match → 2 arms |
| `cli/commands.rs::run_command_metrics` | ABC | **45 → 37** | extracted `resolve_selected_metrics` |
| `py/sarif.rs::collect_offenders` | cognitive / cyclomatic | **31 → ~8 / 15 → 11** | split into `record_threshold_breaches` / `push_child_spaces` |
| `vcs/git/diff_parse.rs::parse_unified_diff` | cognitive / cyclomatic | **20 → 9 / 16 → 9** | extracted `DiffFile::classify_body_line` |
| `cli/markdown_report.rs::language_display_name` | cyclomatic 26 | suppressed | `// bca: suppress(cyclomatic)` — exhaustive wildcard-free `match LANG` display table, kept whole on purpose |

`collapse_scc` (cognitive 17) was left whole: already under the 25
threshold, and a split there would only inflate file-level `nom`/`nargs`.

No public API changes — every modified function keeps its signature; only
private helpers were added.

## Verification

- `make pre-commit` green: fmt, clippy ×2, workspace tests
  `--all-features` (2687 lib + 281 py), doc, udeps, all lint families,
  manpages, self-scan hard + soft tiers, stubtest.
- `.bca-baseline.toml` refreshed (soft tier) — **removal-only** diff
  dropping the 8 now-passing entries; no new offenders.
- Ran `simplify-rust`, `rust-optimize`, `audit-tests` (test-via-revert
  confirmed existing tests guard the macro-ordering path), and a
  high-effort `code-review` — no actionable findings.

Refs #736

### DIFF
--- a/.bca-baseline.toml
+++ b/.bca-baseline.toml
@@ -40,20 +40,6 @@ metric = "nom"
 value = 30.0
 
 [[entry]]
-path = "big-code-analysis-cli/src/lib.rs"
-qualified = "expand_seed_paths"
-start_line = 2758
-metric = "cognitive"
-value = 24.0
-
-[[entry]]
-path = "big-code-analysis-cli/src/markdown_report.rs"
-qualified = "language_display_name"
-start_line = 260
-metric = "cyclomatic"
-value = 26.0
-
-[[entry]]
 path = "big-code-analysis-cli/src/markdown_report/hotspot.rs"
 qualified = "<file>"
 start_line = 1
@@ -101,20 +87,6 @@ qualified = "_native"
 start_line = 829
 metric = "abc"
 value = 67.0
-
-[[entry]]
-path = "big-code-analysis-py/src/sarif.rs"
-qualified = "collect_offenders"
-start_line = 485
-metric = "cognitive"
-value = 31.0
-
-[[entry]]
-path = "big-code-analysis-py/src/sarif.rs"
-qualified = "collect_offenders"
-start_line = 485
-metric = "cyclomatic"
-value = 15.0
 
 [[entry]]
 path = "big-code-analysis-py/src/types_codegen.rs"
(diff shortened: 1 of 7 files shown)


dekobon:
Follow-up (commit 3e19c243): test-quality fixes the refactor surfaced.

The `audit-tests` revert pass found that `preprocess_redefine_after_undef_keeps_macro`
**passed even with `apply_macro_events`' source-order sort removed** — its
`#define`/`#undef`/`#define` fixture is symmetric (ends on a `#define` in
either replay order), so it never exercised the #705 ordering invariant it
documented. Replaced with an asymmetric `#undef FOO` / `#define FOO` fixture
that drops FOO if the sort is missing/reversed, and added a test pinning the
`is_specials` filter that moved into `classify_preproc_node`.

Both verified by test-via-revert: each fails against the perturbed production
line and passes once restored. `make pre-commit` green.

Issue #640, opened 2026-06-10
fix(lib): normalize EOL/trailing newline on in-memory sources so web and Python match the CLI


dekobon:
## Current behavior

The identical 8-byte source `fn f(){}` (no trailing newline) produces different numbers per surface:

- Web `/v1/metrics`: unit space `end_line: 0`, `sloc: 0` — the unit space ends *before its own child function* (the nested fn reports `end_line: 1`).
- CLI on the same bytes in a file: `end_line: 1`, `sloc: 1`.

Cause: the CLI reads via `read_file_with_eol`, which guarantees a trailing `\n` and normalizes CRLF/CR (`src/tools.rs:70-91+`); the web server feeds raw request bytes to the parser (`big-code-analysis-web/src/web/server.rs:263`, `:377`); the Unit space computes `end_line = node.end_row()` un-incremented (`src/spaces.rs:309-316`) on the assumption that a trailing newline exists. CRLF input showed no divergence — only the missing-final-newline case.

Dedup: no existing issue found (searched "trailing newline", `git log --grep=newline`).

## Why it's a problem

REST clients POSTing editor buffers (which commonly lack a final newline) silently get `sloc: 0` and an impossible `end_line: 0`. The same bytes give different numbers depending on surface, breaking the "same canonical numbers everywhere" story (#540/#542). Python `analyze_source` shares the raw-buffer path.

## Proposed fix

Extract the `read_file_with_eol` normalization (trailing-newline guarantee + CRLF/CR normalization) into a shared library helper that also serves in-memory buffers, and call it from the web handlers and Python `analyze_source` before parsing.

## Breaking?

No shape break — value-changing for unterminated/CR-bearing in-memory inputs (snapshot / re-baseline class), bringing them in line with what the CLI already reports for the same content.

Found during the 2.0 UX audit (#505 context), but independently fixable — not gated on 2.0.

---

## Resolution Plan

1. Factor the normalization out of `read_file_with_eol` (`src/tools.rs`) into a helper usable on owned/borrowed byte buffers.
2. Apply it in the web handlers (`server.rs:263`, `:377`) and in Python `analyze_source`.
3. Regression tests: web `/metrics` on the unterminated buffer matches CLI numbers (`end_line: 1`, `sloc: 1`); a Python parity test for the same source; cover CR-only input too.
[...]

PR #1061, opened 2026-07-24
fix(metrics/loc): guard doc-comment row discount at EOF

### LINKED ISSUE #1051: fix(metrics/loc): usize underflow panics on Rust doc comment at EOF
## Summary

`src/metrics/loc/rust.rs:48` performs an unguarded `end - 1` on a `usize`.
When a Rust doc comment is the last thing in the buffer with **no trailing
newline**, `end == start == 0` and the subtraction underflows.

This panics in **both** debug and release profiles, from a 5-byte input,
through the public `analyze()` entry point.

## Reproduction

Standalone crate with a path dependency on this repo (no local edits):

```rust
use big_code_analysis::{analyze, MetricsOptions, Source, LANG};

fn main() {
    let _ = analyze(
        Source::new(LANG::Rust, b"/// x"),
        MetricsOptions::default(),
    );
}
```

| input | debug | release |
|---|---|---|
| `/// x` | PANIC `attempt to subtract with overflow` | PANIC `Hash table capacity overflow` |
| `//! x` | PANIC | PANIC |
| `let x = 1; /// d` | PANIC | PANIC |
| `/// x\n` (control) | ok | ok |
| `// x` (control, not a doc comment) | ok | ok |

Debug panics at `src/metrics/loc/rust.rs:48:21`. Release does **not** save
us: `end` wraps to `usize::MAX`, flows into `add_cloc_lines`
(`src/metrics/loc/shared.rs:46`, itself an unguarded `end - start`), and
then `add_only_comment_lines` extends a `HashSet` over a near-`usize::MAX`
range, panicking inside hashbrown.

## Root cause

```rust
// src/metrics/loc/rust.rs:44-51
// Exclude the last line for `LineComment` containing a `DocComment`,
// since the `DocComment` includes the newline,
// as explained here: https://github.com/tree-sitter/tree-sitter-rust/blob/2eaf126.../src/scanner.c#L194-L195
let end = if node.is_child(DocComment as u16) {
    end - 1
} else {
    end
};
```

The cited invariant — "the `DocComment` includes the newline" — does not
hold at EOF, where the scanner terminates without consuming a newline. The
control rows above isolate this precisely: `/// x\n` is fine, `/// x` is not.

## Why no existing test catches it

Every path that reaches `Loc::compute` in the CLI, web server, and Python
bindings is shielded *incidentally* by `normalize_line_endings`
(`src/tools.rs:545`), which appends a trailing `\n`. The test helpers do the
same (`src/tools.rs:754`, `src/metrics/loc.rs:769`), so no test input can
reach the underflowing branch.

`Source::new` / `Source::from_bytes` (`src/spaces/source.rs:20,35`) perform
no normalization and no validation. `analyze()` is the documented entry
point, and its own rustdoc example (`src/spaces/compute.rs:178-188`)
uses exactly this unshielded path.

## Impact

- **Published API**: any crates.io consumer calling `analyze()` /
  `Ast::parse()` on unnormalized bytes can be panicked by a 5-byte input.
  This is the shape a `cargo-fuzz` harness or any downstream integration
  would naturally use.
- **First-party binaries are not currently affected**, since they all
  normalize first — but that is an accident of the call path, not a
  documented contract, and one refactor away from becoming reachable.

## Suggested fix

**Status: fixed** — see the Resolution section below. Retained for context.

Guard the subtraction and decide the metric semantics deliberately — a doc
comment at EOF with no trailing newline *is* a comment line, so the
"exclude the last line" adjustment should simply not apply.

## Regression test

Add a test that asserts `analyze(Source::new(LANG::Rust, b"/// x"), ..)`
returns `Ok` — constructed through `Source` directly, **not** through a
normalizing helper, or it will pass vacuously.

---

## Resolution

Fixed in `fix(metrics/loc): guard doc-comment row discount`.

**The investigation refined the diagnosis in one important way: the two
panic sites are not two bugs, and the fix I originally suggested here was
right for the wrong reason.**

`init()` returns `(node.start_row(), node.end_row())`, and tree-sitter
guarantees `end_row >= start_row`. All 25 `add_cloc_lines` call sites pass
that span through unmodified — *except* Rust's `LineComment` arm, the only
one that adjusts `end` itself. So that single `end - 1` is the sole source
of both observed panics:

| input shape | rows | where it lands |
|---|---|---|
| `/// x` | `start == end == 0` | underflows at `rust.rs` directly |
| `fn f(){}\n/// x` | `start == end > 0` | subtraction succeeds, drives `end` **below** `start`, underflows `shared.rs:46` |

A guard of `end > 0` would have fixed only the first row of that table. The
correct guard is **`end > start`** — discount the row only when the node
actually spans one, which is also the exact semantics of "exclude the last
line".

Consequently `shared.rs:46` did *not* need to be made non-underflowing: it
has a genuine `end >= start` precondition that every caller satisfies once
`rust.rs` is fixed. It gained a `debug_assert` documenting that precondition
instead, so a future caller reintroducing the inversion fails loudly in
tests rather than wrapping to `usize::MAX` in release and surfacing far away
as a hash-table capacity overflow.

### Sibling sweep

`end - 1` appears in no other `loc` module — the doc-comment newline
adjustment is unique to Rust. Every other `- 1` hit under `src/metrics/` is
inside a test fixture's *source string* or unrelated `cyclomatic()`
arithmetic.

### Verification

- All four panicking inputs analyse cleanly in **both** debug and release,
  with identical results.
- `/// x`, `//! x`, and `// plain` now produce identical LOC — a doc comment
  at EOF counts exactly like a plain comment at EOF.
- **Test-via-revert** per `.claude/rules/testing.md`: with the guard removed
  by a precise inverse edit, the two bug-pinning tests fail with the expected
  panic. The third test (the no-regression guard) passes against both
  versions by design — it exists to prove the guard did not become a no-op.
- `make pre-commit` green: 3,064 lib tests, clippy clean on both feature
  flavours, **no snapshot drift and no baseline refresh needed**.

### No metric drift

For any newline-terminated input the node spans `start + 1`, so `end > start`
holds and the subtraction still happens. Behaviour is bit-identical to before
for every input that did not previously panic.

dekobon:
## Problem

A Rust doc comment ending at EOF without a trailing newline panicked the
analyzer, from inputs as small as `/// x`:

| input | debug | release |
|---|---|---|
| `/// x` | `attempt to subtract with overflow` | `Hash table capacity overflow` |
| `//! x` | panic | panic |
| `fn f(){}\n/// x` | panic | panic |
| `let x = 1; /// d` | panic | panic |
| `/// x\n` (control) | ok | ok |

Release builds do not save us: `end` wraps to `usize::MAX`, flows into
`add_only_comment_lines`, and blows up inside hashbrown while extending a
near-`usize::MAX` range.

Reachable from `analyze` / `Source` directly — the documented library entry
point, whose own rustdoc example uses exactly that path — **and from the
Python `Ast.parse(...).metrics()` fast path**, which hands raw bytes to
`Source::from_bytes` with no `normalize_eol`, unlike `analyze_source` and
`Ast.from_path` (`big-code-analysis-py/src/ast.rs:123,131`). The `bca` CLI
and every `bca-web` endpoint that computes metrics normalize their input
and were unaffected.

## Root cause

`Loc` discounts the extra row a `DocComment` spans, because the
tree-sitter-rust scanner consumes the trailing newline as part of the
token. At EOF there is no newline left to consume, so the node ends on its
own start row and the unconditional `end - 1` underflows.

**The two reported panic sites are one bug, and that distinction drove the
fix.** `init()` returns `(start_row, end_row)` and tree-sitter guarantees
the span is non-decreasing, so all 25 `add_cloc_lines` call sites are safe
— except Rust's `LineComment` arm, the only one that adjusts `end` itself:

| shape | rows | lands at |
|---|---|---|
| `/// x` | `start == end == 0` | underflows in `loc/rust.rs` directly |
| `fn f(){}\n/// x` | `start == end > 0` | subtraction succeeds, drives `end` **below** `start`, underflows `loc/shared.rs` |

A guard of `end > 0` — the obvious reading of the debug panic — fixes only
the first row. The correct guard is **`end > start`**, which is also
exactly the semantics of "exclude the last line": discount the row only
when the node really spans one.

## Changes

- **`src/metrics/loc/rust.rs`** — add the `end > start` guard, ordered
  cheap-operand-first so it short-circuits `is_child`'s FFI child walk on
  plain line comments. The comment records the EOF exception and cites
  `process_line_doc_content` by name rather than a line range, which was
  already one off against the pinned tree-sitter-rust 0.24.2.
- **`src/metrics/loc/shared.rs`** — `add_cloc_lines` gains a
  `debug_assert!(end >= start)` and a documented precondition. It needed no
  behavioural change: the precondition is real and every caller satisfies
  it once `rust.rs` is fixed. Saturating arithmetic here would have traded
  wrong crashes for silently wrong comment counts, which is worse in an
  analysis tool. The comment states the assert's **narrow** scope: it fires
  only for `end < start`, only in debug, and cannot catch a caller whose own
  arithmetic already wrapped — `(0, usize::MAX)` satisfies `end >= start`.
- **`src/metrics/loc.rs`** — four tests plus a `loc_verbatim(lang, source)`
  helper and its Rust binding `rust_loc`.
- **`docs/development/lessons_learned.md`** — records the test-helper trap
  as a sub-example under lesson 7 (see below).

## This changes metric values

Only the **row-0** shapes panicked. When the EOF doc comment sat on any
**later** row, release builds did not crash — they silently reported one
`cloc` too few, which also shifted `blank` and MI's comment percentage.
Measured against pre-fix code under release arithmetic:

| input | pre-fix (release) | post-fix |
|---|---|---|
| `/// x`, `//! x`, `let x = 1; /// d` | **panic** | `cloc=1` |
| `fn f(){}\n/// x` | `cloc=0` *(no panic)* | `cloc=1` |
| `fn f(){}\r\n/// x` | `cloc=0` *(no panic)* | `cloc=1` |
| `/// a\n/// b` (EOF) | `cloc=1` *(no panic)* | `cloc=2` |
| any newline-terminated input | identical | identical |

So any Rust file whose last line is an unterminated doc comment now reports
one more `cloc` (and one fewer `blank`) than a `2.0.1` release build did.
Downstreams should re-check `.bca-baseline.toml` entries and thresholds.

No in-tree snapshots move — every fixture is newline-terminated — confirmed
by a clean `git status` after the full gate.

## Testing

The tests deliberately bypass `check_metrics`. It routes through
`tools::check_func_space`, which does `trim_end().trim_matches('\n')` then
`push(b'\n')` — it strips trailing newlines and appends one, making "a node
ending at EOF" structurally unreachable. The integration suites bypass that
helper but hit the same wall — `read_file_with_eol` also appends a trailing
newline. **No test written the ordinary way could have caught this, and a
regression test added that way would have passed against the unfixed
code.** `loc_verbatim` goes through `analyze(Source::new(..))` with no
normalization, takes a `LANG` (the untestable-at-EOF class is shared by
every `Loc` submodule), and carries a doc comment explaining why it must not
be "simplified" back.

- `rust_doc_comment_at_eof_does_not_underflow` — all five shapes, including
  `/// a\n/// b` at EOF, which release under-counted as 1.
- `rust_doc_comment_at_eof_matches_plain_comment` — parity with `// x`,
  plus absolute pins on the baseline so it cannot certify a wrong value on
  both sides at once.
- `rust_doc_comment_with_trailing_newline_still_discounts_the_row` — the
  no-regression guard, proving the new condition did not become a no-op.
- `rust_doc_comment_crlf_still_discounts_the_row` — the boundary where the
  discount is still owed (`\r` is ordinary content, so the `\n` is still
  consumed), plus the lone-trailing-`\r` case where it is not.

**Test-via-revert** per `.claude/rules/testing.md`, using a precise inverse
edit rather than a file-level restore (which would have wiped the
uncommitted fix): with the guard removed, the bug-pinning tests fail with
the expected panic, while the no-regression tests pass against both
versions — which is their job.

## Sibling sweep

`end - 1` appears in no other `loc` module; the doc-comment newline
adjustment is unique to Rust. Every other `- 1` under `src/metrics/` is
inside a test fixture's source string or unrelated `cyclomatic()`
arithmetic.

## Lessons learned

Recorded as a sub-example under **lesson 7**, not as a new lesson — #7
already states the general pattern (an intermediate stage that "filters,
normalizes, or short-circuits the input") and its prescription. What this
case adds is severity: #7's existing example is a filter in one test's call
path, whereas here two independent chokepoints (`check_func_space` for unit
tests, `read_file_with_eol` for the integration suites) put the input class
out of reach of the whole suite. Lesson numbering is unchanged, so
index-based references from other skills still resolve.

## Validation

`make pre-commit` green — 3,065 lib tests, clippy clean on default and
`--all-features`, markdown lint, self-scan threshold gates, no snapshot
drift, no baseline refresh.

A high-effort code review over this branch found 10 issues, all fixed in
`6b30fca0` and `a0ae4398`; three were factual errors in this description,
corrected above. See the review comment below for the full list.

Fixes #1051

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,22 @@ for historical reference.
 
 ### Fixed
 
+- A Rust doc comment ending at EOF without a trailing newline no longer
+  panics the analyzer (#1051). `Loc` discounts the row that a
+  `DocComment`'s scanner consumes along with its newline, but at EOF
+  there is no newline left to consume, so the node ends on its own start
+  row and the unconditional `end - 1` underflowed. Debug builds panicked
+  at the subtraction; release builds wrapped to `usize::MAX` and
+  surfaced far away as a hash-table capacity overflow while inserting
+  comment rows. Inputs as small as `/// x` were affected. The first-party
+  binaries never reached this — `bca`, `bca-web`, and the Python bindings
+  all normalize line endings, which appends a trailing newline — but
+  library callers using `analyze` / `Source` directly (the documented
+  entry point, and the shape a fuzz harness would use) did. `Loc` now
+  discounts the row only when the node actually spans one, and
+  `add_cloc_lines` carries a `debug_assert` documenting its
+  `end >= start` precondition so a future caller cannot reintroduce the
+  wrap silently. Metric values for all other inputs are unchanged.
 - docs.rs now publishes the **complete** API reference. The published
   build previously used default features only, silently dropping the
   entire feature-gated `vcs` module (change-history metrics, #328) from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,21 +57,29 @@ for historical reference.
 ### Fixed
 
 - A Rust doc comment ending at EOF without a trailing newline no longer
-  panics the analyzer (#1051). `Loc` discounts the row that a
+  crashes or miscounts (#1051). `Loc` discounts the row that a
   `DocComment`'s scanner consumes along with its newline, but at EOF
   there is no newline left to consume, so the node ends on its own start
-  row and the unconditional `end - 1` underflowed. Debug builds panicked
-  at the subtraction; release builds wrapped to `usize::MAX` and
-  surfaced far away as a hash-table capacity overflow while inserting
-  comment rows. Inputs as small as `/// x` were affected. The first-party
-  binaries never reached this — `bca`, `bca-web`, and the Python bindings
-  all normalize line endings, which appends a trailing newline — but
-  library callers using `analyze` / `Source` directly (the documented
-  entry point, and the shape a fuzz harness would use) did. `Loc` now
-  discounts the row only when the node actually spans one, and
-  `add_cloc_lines` carries a `debug_assert` documenting its
-  `end >= start` precondition so a future caller cannot reintroduce the
-  wrap silently. Metric values for all other inputs are unchanged.
+  row and the row was discounted anyway. The symptom split by where the
+  comment sat:
+  - **On the first row** (`/// x` as the whole file): the analyzer
+    panicked — in debug at the subtraction, in release as a hash-table
+    capacity overflow while inserting comment rows.
+  - **On any later row** (`fn f() {}\n/// x`): release builds did **not**
+    crash. They silently reported `cloc` one too low, which also shifted
+    `blank` and the Maintainability Index's comment percentage.
+
+  **This changes metric values.** Any Rust file whose last line is a doc
+  comment with no trailing newline now reports one more `cloc` (and one
+  fewer `blank`) than a `2.0.1` release build did; two such comments
+  shift by two. Re-check `.bca-baseline.toml` entries and thresholds for
+  affected files.
+
+  Reachable from `analyze` / `Source` (the documented library entry
+  point) and from the Python `Ast.parse(...).metrics()` fast path, which
+  — unlike `analyze_source` and `Ast.from_path` — does not normalize
+  line endings. The `bca` CLI and every `bca-web` endpoint that computes
+  metrics normalize their input and were unaffected.
 - docs.rs now publishes the **complete** API reference. The published
   build previously used default features only, silently dropping the
   entire feature-gated `vcs` module (change-history metrics, #328) from
--- a/docs/development/lessons_learned.md
+++ b/docs/development/lessons_learned.md
@@ -422,17 +422,21 @@ spans when its scanner consumes the trailing newline; at EOF there is
 none to consume, so the unconditional `end - 1` underflowed —
 panicking in debug, and in release wrapping to `usize::MAX` and
 surfacing far away as a hash-table capacity overflow while inserting
-comment rows. Every LOC test routes through `check_metrics` →
+comment rows. The unit LOC tests route through `check_metrics` →
 `tools::check_func_space`, which does `trim_end().trim_matches('\n')`
 and then `push(b'\n')`: it strips trailing newlines and appends one.
-"A node ending at EOF" is therefore unreachable through the helper, so
-the suite had no path to the bug — and a regression test written the
-ordinary way would have passed against the unfixed code. The fix added
-a `loc_verbatim` helper calling `analyze(Source::new(..))` with no
-normalization, carrying a doc comment explaining why it must not be
-"simplified" back to `check_metrics`. Sharper than the NaN case above:
-there the filter sat in one test's call path; here it is in the
-canonical helper an entire metric family shares.
+The integration snapshot suites (`tests/serde_test.rs` and siblings)
+bypass that helper entirely — but reach the same wall, because
+`read_file_with_eol` → `normalize_line_endings` ends with an
+unconditional `data.push(b'\n')`. "A node ending at EOF" is therefore
+unreachable from *either* harness, so the suite had no path to the bug
+— and a regression test written the ordinary way would have passed
+against the unfixed code. The fix added a `loc_verbatim` helper calling
+`analyze(Source::new(..))` with no normalization, carrying a doc
+comment explaining why it must not be "simplified" back to
+`check_metrics`. Sharper than the NaN case above: there the filter sat
+in one test's call path; here two independent chokepoints normalize the
+input class out of reach of the whole suite.
 
 **Section-presence tests with no value assertions** (`df84dd27`).
 `wmc_section_present_with_class_summaries`,
--- a/docs/development/lessons_learned.md
+++ b/docs/development/lessons_learned.md
@@ -416,6 +416,24 @@ intermediate stage filters, normalizes, or short-circuits the
 input. Always pair end-to-end tests with direct unit tests on the
 function whose contract is being verified.
 
+**A normalizing shared helper made an entire input class untestable**
+(#1051, `aaddda98`). `Loc` discounts the extra row a Rust `DocComment`
+spans when its scanner consumes the trailing newline; at EOF there is
+none to consume, so the unconditional `end - 1` underflowed —
+panicking in debug, and in release wrapping to `usize::MAX` and
+surfacing far away as a hash-table capacity overflow while inserting
+comment rows. Every LOC test routes through `check_metrics` →
+`tools::check_func_space`, which does `trim_end().trim_matches('\n')`
+and then `push(b'\n')`: it strips trailing newlines and appends one.
+"A node ending at EOF" is therefore unreachable through the helper, so
+the suite had no path to the bug — and a regression test written the
+ordinary way would have passed against the unfixed code. The fix added
+a `loc_verbatim` helper calling `analyze(Source::new(..))` with no
+normalization, carrying a doc comment explaining why it must not be
+"simplified" back to `check_metrics`. Sharper than the NaN case above:
+there the filter sat in one test's call path; here it is in the
+canonical helper an entire metric family shares.
+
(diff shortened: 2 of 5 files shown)


dekobon:
Ran a high-effort code review (8 finder angles) over the branch. It found
10 issues, all now fixed in `6b30fca0` and `a0ae4398`. Three were factual
errors **in this PR's own description**, so the body above has been
corrected — flagging that here since it changes the stated impact.

## The two that matter

**1. The Python bindings were exposed; I said they weren't.**
`PyAst::parse` (`big-code-analysis-py/src/ast.rs:123,131`) calls
`extract_source_bytes` then `Source::from_bytes` with **no** `normalize_eol`
— unlike `analyze_source` (`analysis.rs:422`) and `Ast.from_path`. So
`Ast.parse("/// x", "rust").metrics()` hit the same crash on a release
wheel. The original claim that all three first-party surfaces normalized
was wrong.

**2. "No metric drift" was wrong.** I verified the panic on three inputs
and generalized. Testing a fourth showed release builds do **not** panic
when the doc comment sits past row 0 — they silently under-count `cloc`:

| input | pre-fix (release) | post-fix |
|---|---|---|
| `/// x`, `//! x`, `let x = 1; /// d` | panic | `cloc=1` |
| `fn f(){}\n/// x` | `cloc=0` *(no panic)* | `cloc=1` |
| `/// a\n/// b` (EOF) | `cloc=1` *(no panic)* | `cloc=2` |
| newline-terminated | identical | identical |

Three input classes change value. The changelog now says so and tells
readers to re-check baselines. No in-tree snapshots move (every fixture is
newline-terminated), which is why the gate stayed green and hid this.

## Also fixed

- **The `debug_assert` didn't guard what its comment claimed.** For `/// x`
  the caller's arithmetic wraps *before* the call, so `add_cloc_lines` sees
  `(0, usize::MAX)`, `end >= start` holds, and it passes — then
  `add_only_comment_lines(0, usize::MAX)` overflows anyway. Release disables
  `debug-assertions` *and* `overflow-checks`, so it cannot run there at all.
  Comment now states its actual narrow scope rather than overclaiming.
- **`bca-web` normalization claim was blanket but isn't** — three of seven
  source-consuming handlers deliberately skip it (#640); the conclusion only
  survives because none compute Loc.
- **CRLF was untested** — the one boundary where the discount must still
  apply. Now pinned, including a lone trailing `\r`.
- **`loc_verbatim` hardcoded `LANG::Rust`** behind a language-agnostic name;
  now takes a `LANG`, with `rust_loc` as the explicit binding.
- **Guard operands were ordered expensive-first**; `end > start` is false for
  every plain line comment, so it now short-circuits `is_child`'s FFI walk.
- **Stale `scanner.c#L194-L195` citation** (one off against the pinned
  0.24.2) replaced with the function name, which can't rot.
- **Lessons-learned over-claimed** "every LOC test routes through
  `check_metrics`" — the integration suites use `read_file_with_eol`. They
  hit the same wall, which makes the lesson stronger: two independent
  chokepoints, not one.
- **`assert_eq!(ploc(), 0)` asserted the default value**; added absolute
  baseline pins so the parity test can't certify a wrong value on both sides.

## Not fixed — needs its own issue

`Sloc::sloc()` computes `end - start` with no `+1` for unit spans, so
`analyze(Source::new(LANG::Rust, b"fn main() {}"))` reports `sloc == 0` for
a one-line file, and `mi::inputs_are_empty()` then returns `0.0` for real
input. Pre-existing, affects every language, and correcting it would move
metric values broadly — not something to slip into a doc-comment fix.

`make pre-commit` green: 3,065 lib tests, clippy clean on both feature
flavours, no snapshot drift, no baseline refresh.

Issue #1088, opened 2026-07-27
perf(metrics): thread the ancestor chain through the four remaining Node::parent climbs


dekobon:
> **Resolved.** All four sites are done, across `597a7720`
> (`perf(node): scan a node's children with a cursor`) and `b5a28b2d`
> (`perf(metrics): thread the chain through the last climbs`). Three
> were threaded as proposed below; Elixir's `Npm` / `Npa` was fixed by
> *deleting* the classifier call, whose answer the `defmodule` check
> below it always discarded. The `nom/nested-arrow` probe added as the
> guard went from `k = 1.97` to `k = 1.03`. The decisive cause was not
> in this list: `Node::wraps_any`'s `next_sibling()` child scan was
> itself `O(children × depth)`. See the closing comment for detail.
> Remaining per-node `node.parent()` calls are tracked in #1096.

---

## Summary

[#1084][parent-walk] gave the metric walk an ancestor chain
(`Ancestors` in `src/node.rs`) and threaded it through the three hot
paths its benchmark probes cover: `Checker::is_else_if`,
`Loc::compute`'s `Node::count_specific_ancestors` gate, and Elixir's
`elixir_is_inside_quote_block` (via `Checker::is_func*_with_code`,
`Getter::get_space_kind_with_code`, and `Nom::compute`). Those three
probes moved from `QUADRATIC_BOUND` to `LINEAR_BOUND` in
`big-code-analysis-bench/src/shapes.rs`.

Four call sites were deliberately left on `Ancestors::unknown()`, which
climbs with `Node::parent` and so is still `O(depth)` per step. Each is
correct — `unknown()` answers identically, it just costs more — but each
is a latent quadratic on a deeply nested file, for the same reason
#1084 described.

[parent-walk]: https://github.com/dekobon/big-code-analysis/issues/1084

## The remaining sites

| Site | Reached from | Why it was deferred |
|---|---|---|
| `js_ancestor_walk!` in `src/checker.rs` (`check_if_func!` / `check_if_arrow_func!`) | `Checker::is_func` / `is_closure` for JS / MozJS / TS / TSX | `is_func` / `is_closure` are called from `Nargs`, `Npm`, `function.rs`, and `ops.rs` as well as the metric walk. Only the metric walk holds a chain, so threading one means widening those predicates and every caller. |
[...]

Issue #1084, opened 2026-07-26
perf(metrics): Node::parent lookups make three walks quadratic in nesting depth


dekobon:
## Summary

The benchmark harness added in #1068 measures three metric walks as
quadratic in nesting depth. All three have the same root cause:
`tree_sitter` stores no parent pointer, so `Node::parent` resolves by
descending from the root and is itself `O(depth)`. Any predicate in the
walk that asks a node for its parent is `O(depth)` per node and
`O(depth^2)` over a deeply nested file — however few parent steps it
takes.

This is the same shape as #1052 (`tokens`' per-leaf ancestor walk) and
#1062 (`cognitive`'s `get_nesting_from_map`), both of which were fixed
by propagating state down the traversal instead of recovering it
upwards. These three were not.

## Measured

`cargo bench -p big-code-analysis-bench --bench scaling`, release
build, idle host. Fitted exponent of `time ~ depth^k`:

| Probe | k | Call |
|---|---|---|
| `cognitive/nested-if` | 1.97 | `Checker::is_else_if` |
| `loc/nested-declaration` | 1.95 | `Node::count_specific_ancestors` |
| `nom/nested-quote` | 2.01 | `elixir_is_inside_quote_block` |

Each has a control probe in the same run that isolates the cause:

| Control | k | Difference from the probe above |
|---|---|---|
| `cognitive/nested-while` | 1.01 | identical nesting, no `is_else_if` |
| `loc/nested-while` | 0.99 | identical nesting, no `declaration` node |

So the quadratic term is attributable to the parent lookup, not to
nesting in general.

Absolute cost, for scale: `nom` on 30 KB of nested Elixir `quote`
blocks (depth 1000) takes ~260 ms; `loc` on 21 KB of nested C
declarations takes ~62 ms. Both are ~35x and ~48x the per-byte cost of
their linear controls at the same depth.

## Scope

`Checker::is_else_if` has an impl in 13 languages, so
`cognitive/nested-if` is not a C-only finding.
`Node::count_specific_ancestors` is reached from `loc` (C, C++, Mozcpp,
ObjC, Java, C#, Go, Groovy), `checker.rs`'s JS `check_if_func`,
`checker/python.rs`, `checker/php.rs`, and `cognitive/python.rs`.

## Direction

The #1052 / #1062 fixes both worked by seeding each child's state from
its parent's during the downward walk, which turns an `O(depth)` lookup
into `O(1)`. The same approach should apply here: the walker already
[...]

PR #1095, opened 2026-07-28
perf(cognitive): read function depth off the ancestor chain

### LINKED ISSUE #1062: perf(metrics/cognitive): O(nodes x depth) parent() lookup per node
## Summary

With the `tokens` ancestor walk fixed in #1052, `cognitive` is now the sole
remaining superlinear term in the metrics walk. It calls `Node::parent()`
once per node, and `parent()` is `O(depth)` in tree-sitter, so the metric
costs `O(nodes × depth)` — quadratic on linearly-nested input.

## Root cause

`src/metrics/cognitive.rs:254-262`:

```rust
fn get_nesting_from_map(
    node: &Node,
    nesting_map: &HashMap<usize, (usize, usize, usize)>,
) -> (usize, usize, usize) {
    node.parent()
        .and_then(|parent| nesting_map.get(&parent.id()))
        .copied()
        .unwrap_or((0, 0, 0))
}
```

The map is keyed by node id and each node stores its own nesting triple, so
every child recovers its parent's entry via `node.parent()`. The lookup
itself is `O(1)`; the `parent()` call that produces the key is not.

Called once per node from ~20 per-language modules under
`src/metrics/cognitive/` (`rust.rs:25`, `java.rs:25`, `go.rs:25`,
`python.rs:95`, `kotlin.rs:44`, `groovy.rs:25`, `ruby.rs:37`, `tcl.rs:25`,
`irules.rs:25`, the `js_cognitive!` macro at `cognitive.rs:310`, …).

Two secondary ancestor walks exist and are `O(depth²)` each, but fire far
less often: `increment_function_depth` (`cognitive.rs:264-272`) and
`elixir_is_inside_quote_block` (`cognitive.rs:540-548`).

## Measurements

Release binary, `int main(){return ((((…1…)))); }` — nesting depth only.
Per-metric isolation at depth 4000, after the #1052 fix:

| metric | time |
|---|---|
| **cognitive** | **1080 ms** |
| cyclomatic | 6 ms |
| halstead | 5 ms |
| tokens | 5 ms |
| every other metric | ~5 ms |

End-to-end scaling after #1052 (all metrics), showing the residual
quadratic is entirely cognitive:

| depth | input | all metrics | tokens only |
|---|---|---|---|
| 1000 | 2.0 KB | 74 ms | 4 ms |
| 2000 | 4.0 KB | 285 ms | 5 ms |
| 4000 | 8.0 KB | 1122 ms | 6 ms |
| 8000 | 16 KB | 4247 ms | — |

~4× per doubling. Extrapolating, a 64 KB nested file still costs on the
order of a minute, so the `bca-web` exposure described in #1052 is reduced
but not closed.

## Suggested direction

The traversal already visits parents before children and, since #1052,
carries a `Walk` tag down the stack (`src/spaces/compute.rs`). Extending
that tag with the parent's node id would make the lookup `O(1)`:

```rust
struct Walk { level: usize, in_comment: bool, parent_id: usize }
```

`get_nesting_from_map` would then take `parent_id` directly and never call
`parent()`. The cost is a signature change to `Cognitive::compute`, which
is `pub(crate)` (no SemVer impact) but is threaded through ~20 per-language
modules plus the no-op arm in `src/macros/mod.rs:42-51` — mechanical, but
wide enough to deserve its own PR rather than riding along with #1052.

Worth auditing the same pattern while there: `parent()` being `O(depth)` is
easy to misread as `O(1)`, and there are further per-node `parent()` calls
under `src/checker/` and `src/metrics/abc/`.

## Acceptance

- `cognitive` scales linearly in nesting depth.
- Cognitive values are unchanged — the existing per-language tests are the
  guard, since this is purely how the parent's nesting triple is reached.
- A deep-nesting regression test in the shape of
  `tokens_deep_nesting_is_tractable` (`src/metrics/tokens.rs`), which pins
  tractability by formula rather than wall-clock.

---

## Status

**Fixed — all four sites.**

| site | status |
|---|---|
| `cognitive::get_nesting_from_map` — one `parent()` per node | fixed, PR #1064 |
| `Loc::count_specific_ancestors` | fixed by #1084 (takes an `Ancestors`) |
| `elixir_is_inside_quote_block` (ungated) | fixed by #1084 (takes an `Ancestors`) |
| `cognitive::increment_function_depth` | fixed here |

PR #1064 made whole-file analysis of nested parentheses linear: nesting
depths 8000 / 16000 / 32000 / 64000 take 13 / 23 / 45 / 88 ms.
[#1084][parent-walk] then gave the metric walk an explicit ancestor
chain (`Ancestors` in `src/node.rs`), which retired the `loc` and Elixir
sites and left `increment_function_depth` as the only `Node::parent`
climb still on the `cognitive` path. That one now reads the same chain:
the new `cognitive/nested-fn` scaling probe fits `time ~ depth^k` at
2.04 against the climb and 1.21 against the chain, and at depth 4000 the
walk drops from ~150 ms to ~16 ms. Cognitive values are unchanged in
every language.

[parent-walk]: https://github.com/dekobon/big-code-analysis/issues/1084

The `in_function`-flag mechanism sketched in the comments was not
needed: after #1084 all 19 call sites already have an `ancestors` value
in scope, so the change is a helper signature plus one argument per
call.

`Node::parent`'s cost documentation — the last item in the first comment
— landed with #1084 (`src/node.rs`).

Two corrections to the original triage, both in the comments and worth
repeating here:

- `count_specific_ancestors` is **not** a long walk in practice — its
  `stop` predicate fires on the enclosing `compound_statement`, so it
  costs a single `parent()` call per `Declaration`. Same shape as the
  first site.
- The first Elixir measurement was invalid: the generator indented each
  level, so the input itself grew quadratically. Re-measured with linear
  input growth, it was genuinely quadratic and the worst of the three,
  since it runs for every node behind no metric selection.

The `Node::parent` climbs that remain are outside `cognitive` and
outside every probed walk — the JS/TS `is_func` / `is_closure` walk,
Elixir's `Npa` / `Npm` / `get_func_space_name` / suppression-marker
lookups, and the Halstead `get_op_type` getters. They are tracked in
#1088.

dekobon:
## Summary

Closes the last of #1062's four `Node::parent` sites.
`increment_function_depth` asked every function node whether a function
encloses it by climbing with `Node::parent`, which `tree_sitter`
resolves by descending from the root. That is `O(depth)` per step, so
`Cognitive` stayed `O(depth²)` on nested function definitions
everywhere it is called.

The scan now reads the ancestor chain the walker hands down (`Ancestors`,
from #1084), so a step is a slice index. It was deferred out of #1084 to
bound that change's blast radius; all 19 call sites (22 languages,
counting the four the JS-family macro expands to) already had an
`ancestors` value in scope, so this is one helper signature plus one
argument per call — no per-language logic moves.

Code review then turned up two more per-node climbs inside the same
metric that the issue never listed: Kotlin's `when`-default check and
Ruby's `case`-default check each called `node.parent()` once per `else`
node, so a deeply nested `if`/`else` chain stayed quadratic in the very
metric this PR makes linear on nested definitions. Both read the same
chain now, and `cognitive` has no `Node::parent` call left in
production.

The `in_function`-flag mechanism sketched on the issue was not needed,
and neither was the per-language proof it would have required that each
match arm's kind set equals its `stops` set.

## Measured

`cargo bench -p big-code-analysis-bench --bench scaling`, new
`cognitive/nested-fn` probe (Rust, `Metric::Cognitive`, depths
1000 / 2000 / 4000), fitted `time ~ depth^k`:

| build | k | verdict |
|---|---|---|
| `Node::parent` climb (pre-fix body, new probe) | **2.04** | OVER BOUND |
| ancestor chain | **1.21** | ok (bound 1.50) |

At depth 4000 the walk drops from ~150 ms to ~16 ms. No other probe
moved; `nom/nested-fn` on the same shape fits 1.13 and is the metric
control.

The probe is new because `nom/nested-fn` never reached this function —
it selects `Metric::Nom`. Its shape gains one `if` per level so
`cognitive` scores `n(n+1)/2` on it, which moves if function depth stops
being counted, so the value column guards the walk as well as the
timing.

## Correctness

Cognitive values are unchanged in every language. Two tests pin that:

- `cognitive_function_depth_is_inherited_at_depth` — the closed form
  `n(n+1)/2` at depth 1000 for nested Rust `fn`s.
- `function_depth_surcharge_holds_across_languages` — an `if` costs 1 in
  a top-level function and 2 one function deeper, for C, Objective-C,
  Mozcpp, JavaScript, MozJS, TypeScript, TSX, Ruby, Lua, Bash and
  iRules. Those were the languages whose function-depth arm had no test
  of its own; C++, C#, Groovy, Java, Kotlin, Perl, PHP, Python, Rust and
  Tcl already had one. Go is deliberately absent — its stop set is
  `function_declaration` / `method_declaration`, neither of which the
  grammar admits inside a function body, so the surcharge is unreachable
  there and a nested Go function takes the `lambda` path.

Both verified by perturbation: with `*depth += 1` turned into
`*depth += 0`, the cross-language test reports all 11 rows and the
whole-suite run fails exactly 10 tests out of 3,110 — precise rather
than coarse.

The second commit is a test-audit fix: the surcharge table asserted only
the two costs, so a row whose snippet stopped parsing would keep
reporting 1 and 2 off an error-recovery tree (verified — appending
garbage to the C source left the test green). Each row now rejects an
ERROR-node parse first, which matters for the rows leaning on GNU nested
functions, Lua's nested `function` statement and an iRules `proc` inside
a `proc`.

## Baseline

`RubyCode::compute`'s `halstead.effort` crosses its recorded
`.bca-baseline.toml` value twice here — 52,423 → 53,271 on the added
argument, then → 53,802 on the `case`-default fix. Refreshed with
`make self-scan-write-baseline-headroom` in each of those commits, per
the baseline-refresh discipline in `AGENTS.md`; skipping it turns
`make self-scan` red on a clean checkout for everyone, not just the
author.

## Still open elsewhere

The `Node::parent` climbs that remain are all outside `cognitive` and
outside every probed walk: the JS/TS `is_func` / `is_closure` walk,
Elixir's `Npa` / `Npm` / `get_func_space_name` / suppression-marker
lookups, the Halstead `get_op_type` getters, and per-node lookups in
several `loc`, `npa` / `npm`, `cyclomatic` and `checker` arms. That
last group was missing from the inventories in the CHANGELOG and
`benchmarking.md`, which read as exhaustive; both now say otherwise.
All stay tracked in #1088, which has been updated to record that its
deferred `increment_function_depth` item is done.

## Validation

- `make pre-commit` green on the final tree.
- `cargo test --workspace --all-features`: 3,110 + 75 passing, 0 failed.
- Reviewed with the `review`, `rust-optimize`, `audit-tests` and
  `code-review` passes; every finding is folded into the four commits.
  The Kotlin / Ruby fix is covered by five existing tests, confirmed by
  perturbation (swapping in the grandparent fails exactly those five
  and nothing else).

Fixes #1062

### DIFF
--- a/.bca-baseline.toml
+++ b/.bca-baseline.toml
@@ -765,7 +765,7 @@ path = "src/metrics/cognitive/ruby.rs"
 qualified = "RubyCode::compute"
 start_line = 29
 metric = "halstead.effort"
-value = 53271.17757518335
+value = 53802.68593577912
 
 [[entry]]
 path = "src/metrics/loc/c.rs"
--- a/.bca-baseline.toml
+++ b/.bca-baseline.toml
@@ -728,14 +728,14 @@ value = 16.0
 [[entry]]
 path = "src/metrics/cognitive.rs"
 qualified = "tcl_switch_decision_arms"
-start_line = 475
+start_line = 490
 metric = "nargs"
 value = 7.0
 
 [[entry]]
 path = "src/metrics/cognitive.rs"
 qualified = "tcl_switch_decision_arms"
-start_line = 475
+start_line = 490
 metric = "nexits"
 value = 6.0
 
@@ -765,7 +765,7 @@ path = "src/metrics/cognitive/ruby.rs"
 qualified = "RubyCode::compute"
 start_line = 29
 metric = "halstead.effort"
-value = 52423.215599473966
+value = 53271.17757518335
 
 [[entry]]
 path = "src/metrics/loc/c.rs"
@@ -861,7 +861,7 @@ value = 32.0
 [[entry]]
 path = "src/ops.rs"
 qualified = "ops_inner"
-start_line = 216
+start_line = 221
 metric = "halstead.effort"
 value = 58764.00971804493
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -275,23 +275,29 @@ for historical reference.
   grammars whose cognitive impl is a no-op, `preproc` and `ccomment`, now
   build no map at all rather than one entry per AST node.
 
-  `cognitive`'s second `Node::parent` site is gone with it:
+  `cognitive`'s remaining `Node::parent` sites are gone with it.
   `increment_function_depth` asked every function node whether a
   function encloses it by climbing with `node.parent()`, which kept the
-  metric `O(depth²)` on nested definitions across the 19 languages that
-  call it. It now reads the ancestor chain the walker hands down (the #1084
+  metric `O(depth²)` on nested definitions across its 19 call sites (22
+  languages, counting the four the JS-family macro expands to). It now
+  reads the ancestor chain the walker hands down (the #1084
   mechanism, deferred out of that change), and a new
   `cognitive/nested-fn` depth-scaling probe covers it: `time ~ depth^k`
   fits 2.04 against the climb and 1.21 against the chain, and at depth
-  4000 the walk drops from ~150 ms to ~16 ms. Cognitive values are
-  unchanged — the arithmetic is pinned at depth 1000 by
-  `cognitive_function_depth_is_inherited_at_depth`.
+  4000 the walk drops from ~150 ms to ~16 ms. The two remaining
+  per-node climbs inside the metric — Kotlin's `when`-default check and
+  Ruby's `case`-default check, one per `else` node — read the same
+  chain now. Cognitive values are unchanged; the arithmetic is pinned at
+  depth 1000 by `cognitive_function_depth_is_inherited_at_depth` and
+  across languages by
+  `function_depth_surcharge_holds_across_languages`.
 
-  The `Node::parent` climbs that remain are outside `cognitive` and
-  outside the probed walks — five `Ancestors::unknown()` call sites plus
-  the Halstead `get_op_type` getters — and are tracked in #1088.
-  Operators analysing untrusted input should still bound request
-  concurrency and input size.
+  `Node::parent` climbs remain elsewhere in the crate — the five
+  `Ancestors::unknown()` call sites, the Halstead `get_op_type`
+  getters, and per-node lookups in several `loc`, `npa` / `npm`, and
+  `checker` arms — all outside `cognitive` and outside the probed
+  walks, and tracked in #1088. Operators analysing untrusted input
+  should still bound request concurrency and input size.
 
 - Deeply nested source no longer costs quadratic time in the `tokens`
   metric (#1052). `Tokens` decided whether a leaf sat inside a comment by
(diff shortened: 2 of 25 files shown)


dekobon:
## Test coverage follow-up (`07c4ac15`)

Patch coverage was 93.88% — 92 hits, 4 misses, 2 partials — and all six
sat in one place: the failure-reporting branch of
`function_depth_surcharge_holds_across_languages`.

```text
L9840 partial: if (flat_cost, nested_cost) != (1, 2) {
L9841 miss:        wrong.push(format!(
L9842 miss:            "{lang:?}: expected 1 then 2, got …"
L9843 miss:        ));
L9844 partial: }
L9850 miss:    wrong.join("\n"),
```

A hand-rolled diagnostic that runs only when the test fails is
unreachable on a green run by construction, so the fix is to not write
it. The table now builds a `measured` column and compares it against
`expected` with one `assert_eq!`, which keeps the "show every language
at once" property that the accumulator existed for and improves the
failure output — both columns, so you can see which rows are right:

```text
left:  [(C, 1, 1), (Objc, 1, 1), (Mozcpp, 1, 1), …]
right: [(C, 1, 2), (Objc, 1, 2), (Mozcpp, 1, 2), …]
```

### The gap the report could not see

Chasing those lines sent me back to what actually proves this change
correct. `a_known_chain_answers_exactly_what_climbing_answers` pins the
equivalence the whole PR rests on — a known chain answers exactly what
`Node::parent` climbing answers — over C, Java, Python and Elixir
fixtures. None of them contains a function nested inside a function,
Kotlin's `when`/`else`, or Ruby's `case`/`else`: the three shapes this
PR added as consumers. Parity over the machinery is not parity over the
shape a caller asks about. Added a fixture for each, checked against
their AST dumps rather than assumed.

### Fixture rot

`visited > 20` does not keep a fixture honest. A grammar bump that
flattened the nesting a row was added for would leave a large,
clean-parsing tree that exercises nothing, while every parity assertion
kept passing. `assert_parity` now takes the kinds that must appear
*inside another node of the same kind*; flattening the Rust fixture to
two sibling `fn`s trips it:

```text
rust: no `function_item` sits inside another `function_item`, so the
fixture no longer exercises the nesting it was added for
```

`make pre-commit` green.

Issue #1100, opened 2026-07-29
perf(checker): the exclude_tests attribute scan is O(attributes x depth)


dekobon:
## Summary

[#1096][chains] retired the per-node `Node::parent` calls in the
metric bodies and, following #1088's lesson, the `Node::previous_sibling`
calls the fixed predicates reached through. One sibling walk was left
behind because it needs a different fix, not just a chain: the
`exclude_tests` subtree prune.

`Checker::should_skip_subtree` (`src/checker/rust.rs`, the only override)
calls `rust_outer_attr_marks_test`, which walks *backward* across
consecutive `#[...]` siblings:

```rust
let mut sibling = node.previous_sibling();
while let Some(s) = sibling {
    if s.kind_id() != Rust::AttributeItem { break; }
    …
    sibling = s.previous_sibling();
}
```

`ts_node__prev_sibling` opens with `ts_node_parent`, which `tree_sitter`
resolves by descending from the root, so each step is `O(depth)` and the
walk is `O(attributes × depth)` per node. `metrics_inner` calls
`should_skip_subtree` once per node whenever `exclude_tests` is on — which
this repo's own `bca.toml` sets, so `make self-scan` pays it.

## Why it is not a straight port of #1096's fix

`Node::previous_sibling_under` (added in #1096) scans the parent's
children, so it is `O(children)` per step. Substituting it here would make
the walk `O(attributes × children)` — better on a deep tree, *worse* on a
shallow-but-wide one, and a `source_file` with hundreds of items is the
common case. The right shape is probably a single forward scan of the
parent's children that stops at `node` and reads back over the trailing
run of `attribute_item`s: `O(children)` once, independent of how many
attributes there are.

That needs an `Ancestors` parameter on `Checker::should_skip_subtree`
(one override, one call site — `spaces/compute.rs` already holds a valid
chain at that point, though `ancestors` is currently bound just below the
prune and would move above it).

## Guard

No probe covers `exclude_tests`. One would need a Rust shape that is
affine in depth, nests, and puts an attributed item at each level —
`nested_fns` with `#[inline] fn f()` per level would do, selected with
`MetricsOptions::exclude_tests`, except that `Probe` has no field for
non-default options today. Adding one is part of the work.

Measure before and after rather than assuming the class moved (#1088).

[chains]: https://github.com/dekobon/big-code-analysis/issues/1096

## Resolution

Fixed on `fix/issue-1100` (`80b93bcb`). The backward
[...]

Issue #778, opened 2026-06-14
metrics(loc): multi-line string interior rows counted as PLOC in Python but as blank in Perl (cross-language inconsistency)


dekobon:
## Summary

Interior rows of a multi-line string literal are classified inconsistently across languages: Python counts them as PLOC (issue #415: "real code, not blank lines"), while Perl excludes them from PLOC, where `blank = sloc - ploc - cloc` then mislabels those rows as blank lines even though they contain string content.

## Location

- `src/metrics/loc.rs:788-808` (Python `String` arm — inserts interior rows into PLOC)
- `src/metrics/loc.rs:1442-1459` (Perl string-kind no-op arms — interior rows reach neither PLOC nor CLOC)
- `src/metrics/loc.rs:522-531` (`Stats::blank` = `sloc - ploc - only_comment_lines`)
- Tests pinning the divergence: `python_multiline_string_assignment_ploc` (`loc.rs:3178`), `perl_lloc_string_content_excluded_from_ploc` (`loc.rs:5698`)

## Evidence

Python explicitly inserts the interior rows of a non-docstring multi-line string into PLOC:

```rust
// loc.rs:799-808 — "Its interior physical rows are real code, not blank lines (#415)."
if parent.start_row() != start {
    check_comment_ends_on_code_line(stats, start);
    stats.ploc.lines.insert(start);
}
(start + 1..=end).for_each(|line| {
    stats.ploc.lines.insert(line);
});
```

Perl no-ops every string kind (and `StringContent`), so a multi-line string's interior rows are added to neither PLOC nor CLOC:

```rust
// loc.rs:1442-1459
| P::StringSingleQuoted
| P::StringDoubleQuoted
...
| P::StringContent
| ... => {}
```

The Perl test asserts this directly:

```rust
// perl_lloc_string_content_excluded_from_ploc (loc.rs:5698)
// "my $s = \"line1\nline2\nline3\";"  -> ploc == 2  (rows 0 and 2 only)
```

For that 3-line string: `sloc = 3`, `ploc = 2`, `cloc = 0`, so `blank = 3 - 2 - 0 = 1`. Row 1 (`line2`) is reported as a blank line even though it holds string content. The Python #415 fix added the interior-row insertion specifically to avoid this exact "real code reported as blank" outcome — the two language modules now encode opposite definitions for the identical construct.

## Expected Behavior

[...]

Issue #409, opened 2026-05-29
feat(metrics): make cyclomatic '?' (TryExpression) counting configurable for Rust


dekobon:
## Summary

Make the `?` operator's contribution to **cyclomatic complexity**
configurable for Rust. Today every `?` (`TryExpression`) adds +1 to
both standard and modified cyclomatic
(`src/metrics/cyclomatic.rs:398`), mirroring upstream
`rust-code-analysis`. This is *definitionally* defensible but is a
poor signal when cyclomatic is used as a **maintainability gate**,
because `?` is a comprehension aid, not a genuine branch a reader must
trace.

Motivated by #401: `dump_tree_helper` measured cyclomatic 32, but only
~12 of that was real branching (`if`, `match`, loops). The other ~20
were `?` operators in a linear, easy-to-read sequence of fallible
writes. The function tripped the gate for doing nothing structurally
complex.

## Why the model mismatches

McCabe's cyclomatic complexity (1976) predates error-monad sugar. It
serves two purposes that `?` splits apart:

1. **Test-coverage proxy** — minimum independent paths for branch
   coverage. Here counting `?` is *correct*: `expr?` desugars to
   `match { Ok(v) => v, Err(e) => return Err(e.into()) }`, a real CFG
   edge (the error path), and full branch coverage must exercise it.

2. **Comprehension / maintainability proxy** — how hard the code is to
   read and change. Here counting `?` *over-penalizes*: `let x =
   foo()?; let y = bar()?;` reads as a linear "and then" narrative.
   `?` exists precisely to *remove* the cognitive overhead of explicit
   `match`/`if let Err { return }`. Penalizing it pushes authors
   toward `.unwrap()` (banned in this repo) or verbose matches to
   lower a score — a perverse incentive.

This repo uses cyclomatic as a **gate** (`self-scan`,
`bca-thresholds.toml`), i.e. use (2), where counting `?` yields false
positives. It also already computes **cognitive complexity**
(`src/metrics/cognitive.rs`), which correctly ignores `?` — so the
comprehension-accurate signal exists; cyclomatic is just the one the
gates key on.

## Proposal

Add a toggle for the `TryExpression` contribution to cyclomatic
(standard + modified). Candidate surfaces:

- A field on the cyclomatic computation / a `bca-thresholds.toml`
  knob consumed by the self-scan gate (the most valuable consumer).
- A CLI flag (e.g. `--cyclomatic-count-try=<bool>`) and the
[...]

PR #1101, opened 2026-07-30
perf(metrics): retire the per-node Node::parent calls in the metric bodies

### LINKED ISSUE #1096: perf(metrics): retire the per-node Node::parent calls in the Halstead, ABC, and loc bodies
> **Resolved** on `main` in 6d6a73b8, 8bbce330, 647ea211, 7cad37c1.
> All three groups are done, plus two `O(depth)` climbs this inventory
> could not see (`Node::previous_sibling` in the ABC container walkers
> and Kotlin's Halstead interpolation test, and both links of
> `Node::parent_grandparent_match` in Python's `Cyclomatic`). Four
> probes and three shape controls guard the classes; see the closing
> comment for the before/after table. One `previous_sibling` walk is
> deliberately deferred to #1100 — it needs a different fix, not a
> chain.

## Summary

[#1084][parent-walk] gave the metric walk an ancestor chain (`Ancestors`
in `src/node.rs`); [#1062][cognitive-parent] and [#1088][remaining]
finished threading it through every predicate that took an
`Ancestors::unknown()`. What is left is a different shape of the same
defect: per-node `node.parent()` calls inside per-language *bodies*,
where no chain parameter exists yet.

`tree_sitter` stores no parent pointer, so `Node::parent` resolves by
descending from the root and is `O(depth)`. A per-node call is therefore
`O(depth^2)` over a deeply nested file, however few steps it takes.

`rg '\.parent\(\)' src/` is the authority; the groups below are how the
work divides.

### 1. Halstead `get_op_type` (per node, behind `Metric::Halstead`)

`src/getter/{python,rust,cpp,mozcpp,bash,irules}.rs`. These sit on the
same per-node path #1062 just retired for `Cognitive`, and are the
closest in shape to what has already been fixed. `Getter::get_op_type`
takes no chain, so this needs the parameter threaded through
`Halstead::compute` (~20 impls) and `Ops::new` — note `ops_inner` now
maintains a chain of its own after #1088, so the caller side is ready.

### 2. `src/metrics/abc/`

Every language's `compute` opens with a `node.parent()` (`elixir`,
`python`, `csharp`, `groovy`, `rust`, `go`, `ruby`, `tcl`, `irules`,
`mozcpp`, `perl`, …). `Abc::compute` would take an `Ancestors`
parameter; `compute_per_node` already has one in scope.

### 3. `loc` / `npm` / `npa` / `cyclomatic` / `checker` arms

- `src/metrics/loc/{python,lua,kotlin,perl,tcl,irules,elixir}.rs`
- `src/metrics/npm/` and `src/metrics/npa/` — **unblocked**: their
  `compute` already carries an `Ancestors` after #1088, so these are
  a body-only change with no signature work.
- `src/metrics/cyclomatic/elixir.rs`
- `src/checker/{rust,ruby}.rs`

`Loc::compute` already takes an `Ancestors`, so group 3's `loc` arms are
body-only too.

## Guard

No probe covers any of these. Add one per group before the fix and move
it from a quadratic reading to a linear one in the same change, as
#1084 / #1062 / #1088 each did — see the "Adding a probe" section of
[`docs/development/benchmarking.md`](../blob/main/docs/development/benchmarking.md).

## The lesson #1088 paid for

Threading the chain into a predicate is not sufficient on its own.
#1088 threaded `Ancestors` into the JS-family `is_func` / `is_closure`
and the probe stayed at `k = 1.97`: the remaining `O(depth)` term was in
`Node::wraps_any`, which scanned children with `child(0)` +
`next_sibling()`, and a sibling step resolves its parent. Whoever picks
these up should check the primitives each fixed predicate calls, not
only the predicate's own ancestor access, and should measure rather than
assume the class changed.

[parent-walk]: https://github.com/dekobon/big-code-analysis/issues/1084
[cognitive-parent]: https://github.com/dekobon/big-code-analysis/issues/1062
[remaining]: https://github.com/dekobon/big-code-analysis/issues/1088

dekobon:
Closes #1096.

[#1084][parent-walk] gave the metric walk an ancestor chain and
[#1062][cognitive] / [#1088][remaining] threaded it through every
predicate that took an `Ancestors::unknown()`. What was left was the
same defect one layer down: `node.parent()` calls inside per-language
metric *bodies*, where no chain parameter existed. `tree_sitter` stores
no parent pointer, so `Node::parent` descends from the root and is
`O(depth)` — a per-node call is `O(depth²)` over a deeply nested file
however few steps it takes.

Metric values are unchanged for every language. The whole workspace
suite and the integration snapshots pass with no accepted snapshot.

## The three groups

**Halstead `get_op_type`.** `Getter::get_op_type`,
`get_op_type_with_code`, `get_operand_id`, `Halstead::compute`, and
`compute_halstead` take an `Ancestors`, fed from `compute_per_node` and
`ops_inner` — both already maintained a chain. The six impls that read a
parent (`python`, `rust`, `cpp`, `mozcpp`, `bash`, `irules`) index it
now; Python's docstring arm wants a grandparent as well, which
`Ancestors::iter` hands over in two steps.

**`src/metrics/abc/`.** Not shaped the way the issue described. The
`node.parent()` is not at the top of `compute`; it is inside each
language's `<lang>_inspect_container`, which resolves the condition
slot's parent only to seed a boolean-context flag — and every caller had
already descended from that parent. So the parent is passed in rather
than rediscovered. `Abc::compute` takes an `Ancestors` for the arms that
genuinely ask about the *node's* own parent (`<` / `>` disambiguation,
the `&&` / `||` chain walkers).

**`loc` / `npm` / `npa` / `cyclomatic` / `checker`.** `loc`'s seven arms
and the shared `add_multiline_string_ploc` read the chain
`Loc::compute` already carried. `Npm::compute`, `Npa::compute`,
`Cyclomatic::compute` / `compute_with_options`, and
`Checker::is_useful_comment` gained the parameter. `rm_comments` grew a
chain of its own by the metric walk's truncate/push rule, built through
`Ancestors::checked` so a bookkeeping slip fails a debug build rather
than silently feeding Rust's `is_useful_comment` the wrong parent.

## Two climbs the inventory could not see

`rg '\.parent\(\)' src/` is the authority for direct calls, but two
`O(depth)` climbs hide behind helpers whose call sites it does not
match. This is #1088's lesson — *a chain-fed predicate is only as linear
as the primitives it calls* — arriving from the other direction.

- **`Node::previous_sibling`.** All seven ABC container walkers ask it
  whether a ternary's `?` / `:` precedes the operand, and Kotlin's
  Halstead short-interpolation test asks it once per `string_content`
  token and once per operand. `ts_node__prev_sibling` opens with
  `ts_node_parent`, so it carries the identical cost; without this the
  fix would not have held on a ternary shape. New
  `Node::previous_sibling_under` scans the known parent's children
  instead.
- **`Node::parent_grandparent_match`.** Python's `Cyclomatic` calls it
  for every `else` token to ask whether the token opens a `for` /
  `while` / `try` else-clause rather than an `if`, and both links
  climbed.

## Measurements

Four probes and three shape controls, added to
`big-code-analysis-bench/src/shapes.rs` with the fix, measured on the
same idle host either side:

| probe | k before | k after | depth 4000, before → after |
|---|---|---|---|
| `halstead/nested-not` | 1.99 | 0.99 | ~478 ms → ~0.57 ms |
| `abc/nested-if` | 2.00 | 1.14 | ~808 ms → ~3.3 ms |
| `cyclomatic/nested-ternary` | 2.06 | 1.27 | ~1.13 s → ~3.4 ms |
| `loc/nested-quote` | 2.00 | 1.03 | ~9.6 s → ~9.2 ms |
| `halstead/nested-paren` (control) | 0.99 | 0.97 | — |
| `abc/nested-block` (control) | 1.18 | 1.18 | — |
| `cyclomatic/nested-and` (control) | 1.26 | 1.31 | — |

`loc/nested-quote` has no shape control of its own: Elixir's `loc`
catch-all arm fires for every named node, so there is no version of the
shape with the trigger removed. `nom/nested-quote` — the same source
under a metric that does not ask for a parent — stands in.

## Tests

Both new unit tests were verified by perturbing the production line and
confirming each is the **only** failure:

- `rust_keeps_a_macro_token_comment_and_strips_an_ordinary_one` fails
  alone when `rm_comments` hands `is_useful_comment` an empty chain.
- `previous_sibling_under_agrees_with_the_authoritative_lookup` fails
  alone (alongside `java_ternary_conditions`) when the scan returns the
  wrong sibling. Its comment records the one break it cannot see — a
  scan that never finds the node falls back to the authoritative lookup
  and is correct but slow, which is the `abc/nested-if` probe's job.

The existing ABC suite is not vacuous about either change: passing the
wrong node as `parent` fails `java_if_single_conditions` and
`java_bool_returning_terminal_kinds_count`, and stubbing the ternary
sibling seed to `None` fails `java_ternary_conditions`.

## Also here

- `refactor(halstead)`: every language's `Halstead::compute` forwards to
  `compute_halstead` verbatim, so the added parameter grew 23 identical
  impls from three lines to eight each. One macro emits them all
  (−204 lines). No impl carried a per-language comment to hoist.
- `.bca-baseline.toml` refreshed with
  `make self-scan-write-baseline-headroom` in the same commits, per the
  baseline-refresh discipline in `AGENTS.md`.

## Not in scope

One `previous_sibling` walk is deliberately left: the `exclude_tests`
attribute scan in `Checker::should_skip_subtree`. Porting
`previous_sibling_under` to it would trade `O(attributes × depth)` for
`O(attributes × children)`, which is worse on the shallow-but-wide trees
that dominate — it wants a single forward scan instead, plus a probe
that can select `exclude_tests`, which `Probe` cannot express today.
Filed as #1100.

After this, `rg '\.parent\(\)' src/` returns only the `Ancestors`
fallback itself, `Node::get_parent` (dead code), the one-off start node
of a `dump`, `act_on_node`'s chain seed, and tests.

[parent-walk]: https://github.com/dekobon/big-code-analysis/issues/1084
[cognitive]: https://github.com/dekobon/big-code-analysis/issues/1062
[remaining]: https://github.com/dekobon/big-code-analysis/issues/1088

### DIFF
--- a/.bca-baseline.toml
+++ b/.bca-baseline.toml
@@ -98,14 +98,14 @@ value = 7.0
 [[entry]]
 path = "big-code-analysis-cli/src/commands/check.rs"
 qualified = "emit_check_results"
-start_line = 560
+start_line = 564
 metric = "nargs"
 value = 8.0
 
 [[entry]]
 path = "big-code-analysis-cli/src/commands/check.rs"
 qualified = "filter_by_baseline"
-start_line = 371
+start_line = 375
 metric = "nargs"
 value = 8.0
 
@@ -632,35 +632,56 @@ path = "src/getter/bash.rs"
 qualified = "BashCode::get_op_type"
 start_line = 31
 metric = "halstead.effort"
-value = 54681.01625588078
+value = 64170.027458982564
 
 [[entry]]
 path = "src/getter/elixir.rs"
 qualified = "ElixirCode::get_op_type"
 start_line = 142
 metric = "halstead.effort"
-value = 48241.44928566978
+value = 56464.822908007496
+
+[[entry]]
+path = "src/getter/irules.rs"
+qualified = "IrulesCode::get_op_type"
+start_line = 19
+metric = "halstead.effort"
+value = 53163.06867422894
 
 [[entry]]
 path = "src/getter/perl.rs"
 qualified = "PerlCode::get_op_type"
 start_line = 17
 metric = "halstead.effort"
-value = 65994.57830587434
+value = 76538.42828561828
 
 [[entry]]
 path = "src/getter/php.rs"
 qualified = "PhpCode::get_op_type"
 start_line = 31
 metric = "halstead.effort"
-value = 48279.902911067526
+value = 54691.60133906107
+
+[[entry]]
+path = "src/getter/python.rs"
+qualified = "PythonCode::get_op_type"
+start_line = 16
+metric = "halstead.effort"
+value = 48346.66496041094
 
 [[entry]]
 path = "src/getter/ruby.rs"
 qualified = "RubyCode::get_op_type"
 start_line = 21
 metric = "halstead.effort"
-value = 72634.96827369716
+value = 84039.86210807812
+
+[[entry]]
+path = "src/macros/mod.rs"
+qualified = "<file>"
+start_line = 1
+metric = "loc.sloc"
+value = 768.0
 
 [[entry]]
 path = "src/metrics/abc/c.rs"
@@ -672,21 +693,28 @@ value = 15.0
 [[entry]]
 path = "src/metrics/abc/cpp.rs"
 qualified = "CppCode::compute"
-start_line = 105
+start_line = 104
 metric = "cyclomatic"
 value = 15.0
 
 [[entry]]
 path = "src/metrics/abc/go.rs"
 qualified = "GoCode::compute"
-start_line = 95
+start_line = 94
 metric = "cyclomatic"
 value = 15.0
 
 [[entry]]
+path = "src/metrics/abc/go.rs"
+qualified = "GoCode::compute"
+start_line = 94
+metric = "halstead.effort"
+value = 50933.14135135399
+
+[[entry]]
 path = "src/metrics/abc/kotlin.rs"
 qualified = "KotlinCode::compute"
-start_line = 154
+start_line = 156
 metric = "cyclomatic"
 value = 16.0
 
@@ -707,21 +735,28 @@ value = 16.0
 [[entry]]
 path = "src/metrics/abc/perl.rs"
 qualified = "PerlCode::compute"
-start_line = 200
+start_line = 199
 metric = "halstead.effort"
-value = 67862.8313802447
+value = 71163.29612014526
 
 [[entry]]
 path = "src/metrics/abc/perl.rs"
 qualified = "perl_inspect_container"
 start_line = 55
 metric = "halstead.effort"
-value = 53345.08959628053
+value = 48499.6860937902
+
+[[entry]]
+path = "src/metrics/abc/php.rs"
+qualified = "php_count_unary_conditions"
+start_line = 95
+metric = "cognitive"
+value = 25.0
 
 [[entry]]
 path = "src/metrics/abc/rust.rs"
 qualified = "RustCode::compute"
-start_line = 128
+start_line = 127
 metric = "cyclomatic"
 value = 16.0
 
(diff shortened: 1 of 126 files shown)


dekobon:
Coverage follow-up in 74773840.

Patch coverage was 97.04% (25 missing). Reproduced locally with the same
`cargo llvm-cov report --codecov` CI uploads, then split the misses into
what a test can reach and what it cannot.

**Six real gaps, now covered** — all pre-existing, made visible because
this PR touched their signatures:

- `Cyclomatic::compute` for Rust (7 lines, the largest block). The walk
  calls `compute_with_options` directly so it can pass
  `count_cyclomatic_try` through, which left the plain trait entry point
  with no caller anywhere — tests included — while #409 fixes its
  default as "`?` counts".
- `Checker::is_useful_comment` for C, C++, Mozilla C++, and
  Objective-C. Five C-family impls share one Aho-Corasick automaton for
  bindgen's `<div rustbindgen` marker; only `Ccomment` had a test.
- The `add_multiline_string_ploc` call sites for Go, Kotlin, and
  Mozilla C++ — three of nineteen with no test, folded into the
  existing #778 cross-language parity test.
- `csharp_walk_for_statement`'s paren/unary condition slot and both
  halves of `groovy_walk_for_statement`'s child-index split. Every
  existing `for` test in those languages uses `i < n`, whose `LT` token
  arm counts the condition without entering the walker.

Each was verified by perturbation: inverting `RustCode::compute`'s
delegation flag and stubbing C's `is_useful_comment` to `false` each
fail exactly one test and nothing else.

**23 lines remain uncovered**, none of them reachable by a test:

- 13 are llvm-cov artifacts of the signature change — trait method
  *declarations* (`Abc`, `Halstead`, `Npm`), a macro repetition's `)+`,
  a multi-line match-arm guard whose body line *is* covered, and two
  format arguments inside a new test's failure message.
- 6 are defensive branches the dispatcher cannot reach: the ABC ternary
  boolean-context seed in four language families that have no ternary
  walker arm, Python `loc`'s no-parent guard for a `String`, and
  `Getter::get_op_type`'s default body.

The reachability analysis for those is filed as #1102 rather than
papered over — the right fix there is the ternary walker arm #403 left
unfinished, not a unit test that hands the helper a parent production
never produces.

`make pre-commit` is green on the pushed tree.

Issue #1203, opened 2026-08-04
fix(npm): go and rust still enable npm/npa on function spaces


dekobon:
> **Fix open in #1206**, not yet merged — this issue closes when that
> PR lands (its `Fixes #1203` trailer does it automatically). Submodule
> snapshots are already pushed as
> `big-code-analysis-output@a23ea6a8`. Emission becomes a pure function
> of `SpaceKind` for **every** language, not just Go and Rust. Two corrections to the analysis below, both verified by
> measurement — see
> [the closing comment](https://github.com/dekobon/big-code-analysis/issues/1203#issuecomment-5182363247):
>
> 1. The resolution plan's claim that clearing the flag "suppresses that
>    space's block **without** losing the parent roll-up" is false for
>    the shapes this issue reports. Go's root enables only on a direct
>    `MethodDeclaration` child and Rust `npa` only on a module-scope
>    `struct`, so in 4 of the 9 affected `serde` files the root carries
>    no block and a plain clear would have deleted the counts from
>    output rather than relocating them.
> 2. The deviation was wider than Go and Rust: all 18,064 C++
>    `namespace` spaces and 13,703 file roots in the corpora were member
>    scopes emitting nothing.
>
> The Go container-`SpaceKind` question in **Scope** was decided
> **no** — it is a space-tree shape change for the next major. Its
> consequence is documented instead: no `npm` / `npa` threshold can fire
> on Go source.

## Summary

#1197 gave `npm` / `npa` a single emission rule —
`SpaceKind::is_member_scope`: containers and the file root carry the
block, a function space never does — and routed ten languages through
`metrics::opens_member_scope` to enforce it.

Go and Rust were deliberately left on their own node-kind gates, on the
grounds that those gates target real containers. For Go that is not
true, and for Rust it is only half true.

## Go

`GoCode::get_space_kind` has no container kind at all — Go spaces are
only `Function` and `Unit`. The enable therefore targets the file root
and, in one path, whatever space encloses a type literal:

- `src/metrics/npm/go.rs:25` enables on `G::SourceFile`, which is
  `SpaceKind::Unit`.
- `src/metrics/npa/go.rs:116` enables on `G::StructType`, which opens no
  space, so the flag lands on the enclosing space. For a struct declared
  inside a function body that is a **`Function` space** — precisely the
  shape #1197 removed everywhere else.
- `src/metrics/npm/go.rs:90` does the same for `G::InterfaceType`; its
[...]

Issue #1200, opened 2026-08-03
fix(nargs): C/C++ pointer/reference-return functions report nargs = 0


dekobon:
> **Status: FIXED** on `fix/1200-c-family-nargs-declarator` (PR pending).
> All shapes below now report their real arity in C, C++, Mozcpp and
> Objective-C, and the nested `int (*fp(…))(…)` case is fixed rather than
> deferred.
>
> **Two corrections to the plan below**, both detailed in the
> [resolution comment](https://github.com/dekobon/big-code-analysis/issues/1200#issuecomment-5185708582):
>
> 1. **Step 1's field-only descent is not sufficient.** `reference_declarator`
>    and `parenthesized_declarator` expose *no fields at all* in any of the
>    four pinned grammars, so walking the `declarator` field repairs the
>    pointer cases and leaves every reference case at 0. The shipped walk
>    falls back to the last named non-comment child, and stops at a node
>    already carrying `parameters` so a C++ lambda is not mis-read.
> 2. **Step 4's `FIXME` is not needed.** Taking the *innermost*
>    `parameters`-bearing link rather than the first fixes the nested
>    function-pointer-return shape for free.
>
> Step 2 was also widened: rather than swapping seven call sites, the three
> byte-identical `CppCode`/`CCode`/`MozcppCode` `compute` overrides — the
> reason one bug existed in triplicate — were replaced by the `params_owner`
> hook. Step 3's sweep found the getter equivalent, filed as #1208.

---

## Summary

C and C++ functions whose return type is a pointer or reference report `nargs.function_args = 0` regardless of their actual parameter count.

Found during 2.1.0 pre-release cross-validation against lizard 1.23.0: **37 of 471 matched functions (7.9%)** in the DeepSpeech/kenlm C++ corpus under-count to zero.

## Reproduction

```c
FILE *f(int a, int b, int c) { return 0; }   /* nargs = 0, expected 3 */
int *g(int a) { return 0; }                  /* nargs = 0, expected 1 */
```

Same result for `int **`, `int &`, and `static int *` return types, in both C and C++. Verified identical on v2.0.0 and current `main` — this is a long-standing bug, not a 2.1.0 regression.

Real corpus examples (all report 0):
- `PathTrie::get_path_trie` (expected 3)
- `DS_SpeechToTextWithMetadata` (expected 4)
- `BinaryFormat::GrowForSearch` (expected 3)

## Root cause

[...]

Issue #1196, opened 2026-08-03
question(nargs): the threshold gate sums a function's closures' parameters into its own


dekobon:
## Summary

`bca check --threshold nargs=N` gates on `nargs.total()`, which is a
function's own parameter count **plus every nested closure's**. So a
function with three parameters and a two-parameter sort comparator
scores five, and the offender line reads as though it had five
parameters.

The extractor is one line
(`big-code-analysis-cli/src/thresholds.rs`):

```rust
MetricExtractor {
    name: "nargs",
    extract: |m| m.nargs.total() as f64,
    metric: Metric::Nargs,
},
```

`Stats::total()` is `function_args + closure_args` over the space, and a
function space includes its nested closures.

## Measured

Found while working #1183, which proposed converging this repository's
`nargs` limit from 7 to the shipped default of 5. I measured **all 76**
functions that limit would newly gate — not a sample:

| | count |
|---|---|
| would be newly gated at `nargs = 5` | 76 |
| of those, with **6+ parameters of their own** | **17** |
| carrying closure args | 59 |
| with **four or fewer** own parameters | 44 |

The extreme is `Parser<T>::filters` (`src/parser.rs`): **one** parameter
of its own, plus five contributed by closures in its body.

So 78% of the offenders a stricter limit would produce are not
parameter-heavy functions. Acting on them means restructuring or
removing idiomatic inline closures — the opposite of what the metric is
meant to encourage.

## Why this matters beyond one repo

`nargs` is one of the ten thresholds `bca init` scaffolds
(`big-code-analysis-cli/src/default_thresholds.rs`, default 5), so every
adopting project inherits this. A Rust or JavaScript codebase using
iterator chains and callbacks will see offenders whose signatures are
already small, and the offender line gives no hint that the number came
from a closure.

It also distorts the corpus measurement behind #1140: the per-language
97.5th-percentile `nargs` figures were computed from the same summed
value, so languages with heavy closure use are overstated relative to
languages without.

## Question

Should the gate key on `function_args` rather than `nargs.total()`?

Arguments for the current behaviour: a function containing five closures
that each take arguments *is* carrying a lot of interface surface, and
the split is available separately in the JSON output.

Arguments against: the threshold's name and its documentation both
describe "number of arguments" of a function, the offender row prints a
single number with no provenance, and the remediation the number implies
(fewer parameters) is not the remediation that would clear it.

A third option is to keep `total()` but make the offender row show the
[...]

PR #1207, opened 2026-08-04
fix(nargs): stop counting comments as parameters

### LINKED ISSUE #1201: fix(nargs): comments inside parameter lists are counted as parameters
## Summary

Comments inside a parameter list are each counted as a parameter by `nargs`, in every language using the generic `compute_args` path.

Found during 2.1.0 pre-release cross-validation against lizard 1.23.0. Reproduced identically in **C, C++, JavaScript, TypeScript, Python, Rust, and Java**; verified identical on v2.0.0 and current `main` — long-standing, not a 2.1.0 regression.

## Reproduction

```c
int h(int a /* one */, int b /* two */) { return a; }   /* nargs = 4, expected 2 */
```

Real corpus examples (kenlm, comment-documented parameters):
- `DontBhiksha::DontBhiksha` — 4 params + 3 `/*…*/` comments → reports 7
- `ReadBackoff` — 2 params + 1 comment → reports 3

## Root cause

The shared `compute_args` (`src/metrics/nargs.rs:212`) counts every child of the parameter-list node that `Checker::is_non_arg` does not exclude, and every language's `is_non_arg` lists only punctuation (parens/commas). Comment nodes pass the filter.

Note: the **Tcl impl already special-cases comment children** — the generic path never received the same fix. That existing handling is the pattern to generalize (exclude `Checker::is_comment` nodes in `compute_args` itself, rather than adding a comment kind to each language's `is_non_arg` list).

## Impact

Inflates `nargs.function_args` / serialized sums wherever parameters carry inline comments — common in C/C++ codebases that comment out parameter names in declarations-with-bodies (`void f(int /*unused*/)`). Since #1196 the `nargs` gate reads the per-callable own count, so a commented signature can now trip a gate its real arity would pass.

## Regression test requirement

Per-language fixtures with block comments between parameters (at minimum C, C++, JS, TS, Python, Rust, Java — plus Tcl to pin its existing correct behavior against a shared-path refactor). Per `.claude/rules/testing.md`, verify by revert: tests must fail against the current implementation.

---

---

## Status: fixed on `fix/1201-nargs-comment-params` (`08ee00da`), not pushed

Two claims in the scoping below turned out to be wrong; see the
[resolution comment](https://github.com/dekobon/big-code-analysis/issues/1201#issuecomment-5183619503)
for the evidence.

- **Tcl had not already solved this.** `compute_tcl_args` filters
  positively for `Tcl::Argument` and never sees a comment, because Tcl
  has none in an argument list — `proc h {a\n# c\n b}` genuinely
  declares four arguments. Tcl needed no change and is now a guard.
- **`compute_kotlin_lambda_args` was a fourth broken loop** this issue
  does not name. Kotlin lambdas counted comments; Kotlin functions did
  not.

Fixed by one shared `count_args<T>` replacing four negative filters;
`count_elixir_args` is deleted as redundant. Repaired: C, C++, Mozcpp,
Objective-C functions, JavaScript, MozJS, TypeScript, TSX, Python, Rust,
Java, C#, PHP, Ruby, Groovy functions, Elixir, Kotlin lambdas. Swept as
guards: Go, Lua, Objective-C methods and blocks, Kotlin functions,
Groovy closures, Perl.

854 corpus values move, all downward. `make pre-commit` passes; patch
coverage is 100%.

**Remaining before this can close:** push the submodule commit
`89e4c973` to `dekobon/big-code-analysis-output`, push the branch, open
the PR.

## Resolution Plan

Confirmed against `main` (`39674df6`):
`int h(int a /* one */, int b /* two */)` reports `function_args: 4`.
`compute_args` (`src/metrics/nargs.rs:212`) counts every child of the
`parameters` node that `T::is_non_arg` rejects, and every language's
`is_non_arg` lists punctuation only.

1. **Fix the shared path**, not each language's `is_non_arg`. In
   `compute_args`, change the predicate at `:223` to
   `if !T::is_non_arg(n) && !T::is_comment(n)`. `T: Checker` already
   provides `is_comment`, so no trait widening and no per-language edit
   is needed, and every language on the generic path is fixed at once.
   Adding a comment kind to twenty `is_non_arg` lists would be the same
   fix twenty times over, each one able to be forgotten on the next
   language — this is exactly the "fix all affected siblings together"
   case, and the shared function *is* the sibling set.
2. **Sweep the three hand-rolled loops that bypass `compute_args`.**
   `rg -n 'is_non_arg' src/metrics/nargs.rs` finds them:
   - `:581` (Perl) already carries `&& !PerlCode::is_comment(n)` — leave
     it, but note in a comment that it now duplicates the shared rule, so
     the next reader does not delete one believing the other covers it.
   - `:666` `count_elixir_args` has **no** comment filter and is not
     routed through `compute_args`. Same one-line fix; Elixir permits a
     comment inside a parameter list, so add a fixture rather than
     assuming it is unreachable.
   - Tcl's special case (cited in the issue as the pattern to
     generalize) becomes redundant once the shared filter lands. Decide
     deliberately: either drop it and let Tcl ride the shared path, or
     keep it and say why. Whichever is chosen, Tcl needs a fixture
     pinning its *current* correct answer so the refactor is provably a
     no-op for it.
3. **Check the `is_bare_param` early return** (`:217-221`). A bare
   parameter is a single identifier node, never a comment, so that arm
   needs no change — confirm with a fixture rather than by reading, since
   this is the one path the new predicate never runs on.
4. **Fixtures**, one per language, with a block comment between
   parameters and a second with a comment attached to a parameter that
   has no name (`void f(int /*unused*/)`, the real-world C++ shape called
   out in the issue): C, C++, Mozcpp, Objective-C, JavaScript,
   TypeScript, Python, Rust, Java, Elixir, plus Tcl and Perl as
   no-change guards. Assert the exact `function_args` on the function's
   own space. Include a line comment (`//` / `#`) as well as a block
   comment where the language allows one inside a parameter list — the
   two are separate `is_comment` arms in several languages and a block-
   comment-only fixture leaves the other arm untested.
5. **Verify by revert** per `.claude/rules/testing.md`. The fix is a
   two-token predicate change, so revert it and confirm every new fixture
   fails and nothing else does. Per the same rules file's "one input per
   independent path" item, the Elixir and generic-path fixtures must each
   fail on their own — a single shared fixture would leave whichever loop
   it does not reach unproven.
6. **Refresh the integration snapshots.** kenlm's comment-documented
   parameters (`DontBhiksha::DontBhiksha` 7 → 4, `ReadBackoff` 3 → 2)
   move, so `tests/repositories/big-code-analysis-output` churns. All
   four AGENTS.md submodule steps apply, including recording the new
   submodule SHA in the same parent commit.
7. **Document.** `CHANGELOG.md` `## [Unreleased]` → `Fixed`: comments
   inside a parameter list are no longer counted as parameters; serialized
   `nargs` values fall for commented signatures, and a signature that
   previously tripped `bca check --threshold nargs=N` on its comments
   alone will now pass.

**Sequencing note.** This and #1200 both change `nargs` on the same C/C++
corpus and both churn the same submodule snapshots. Landing them in one
branch — or strictly one after the other with a snapshot refresh between —
avoids two rounds of `.snap` conflicts. They move the numbers in opposite
directions (#1200 raises, this lowers), so a combined snapshot diff is
also easier to read for whether each function moved for the right reason.

## Assessment

| Dimension  | Rating |
|------------|--------|
| Difficulty | Low |
| Complexity | Medium |
| Priority   | High |

**Difficulty — Low.** The production fix is two tokens in one shared
function plus the same two in `count_elixir_args`. Everything else is
fixtures.

**Complexity — Medium.** The change itself is isolated to one module, but
it lands on every language at once, needs a decision about Tcl's now-
redundant special case, and churns integration snapshots across the
corpus with a submodule bump.

**Priority — High.** It reports incorrect metric values in at least seven
languages, and since #1196 an inflated count can trip a `nargs` gate the
function's real arity would pass — a false positive in a blocking CI gate.

dekobon:
Fixes #1201.

tree-sitter attaches a comment written inside a parameter list as a
direct child of the parameter-*list* node, not inside the parameter it
documents. Every `nargs` filter listed punctuation only, so each comment
scored one: `int h(int a /* one */, int b /* two */)` reported 4, and
the C++ idiom for a deliberately unused parameter,
`void f(int /*unused*/)`, reported 2.

Since #1196 the gate reads the per-callable own count, so a commented
signature could trip a `bca check --threshold nargs=N` its real arity
would pass.

## The fix

One shared `count_args<T>` replaces four separate negative filters, so
the rule is stated once and the next language inherits it rather than
needing the same edit in twenty `is_non_arg` impls. Elixir's
`count_elixir_args` turned out to be identical to the shared helper once
fixed, and is deleted.

**Repaired:** C, C++, Mozcpp, Objective-C functions, JavaScript, MozJS,
TypeScript, TSX, Python, Rust, Java, C#, PHP, Ruby, Groovy functions,
Elixir, Kotlin lambdas.

**Verified already correct, swept as regression guards:** Go, Lua,
Objective-C methods and blocks, Kotlin functions, Groovy closures, Perl.
Bash has no formal parameter list.

## Two corrections to the issue's scoping

Both established by AST dump rather than by reading the code.

**Tcl had not "already solved this" — it has no problem to solve.** The
issue proposed generalising Tcl's handling; `compute_tcl_args` filters
*positively* for `Tcl::Argument` and never encounters a comment, because
Tcl recognises one only where a command is expected. So

```tcl
proc h {a
  # c
  b} { return $a }
```

genuinely declares four arguments named `a`, `#`, `c` and `b`, and the
grammar emits four `argument` nodes with no comment node. Tcl's reported
4 is correct and unchanged, now pinned by
`tcl_has_no_comment_inside_a_parameter_list`.

**`compute_kotlin_lambda_args` was a fourth broken loop the issue does
not name.** It filtered `!= Kotlin::COMMA`, so
`{ a: Int, /* one */ b: Int -> a }` reported 3. Kotlin *functions* were
fine (positive `Parameter` filter), which is why a function-only check
would have missed it.

## Verification

- **Per-loop revert.** Before the loops were consolidated each was
  reverted independently and failed only its own fixtures. After
  consolidation there is one line to revert; doing so fails 37 of 43
  fixtures, and the 6 survivors are exactly the positive-filter guards.
- **The guards are live, not decorative.** Swapping Go's positive
  `parameter_declaration` filter for a negative one fails both Go rows
  and nothing else.
- **Kotlin widening is safe.** `{ (a, b), c -> }` = 2, `{ (a, b) -> }` =
  1, `{ -> }` = 0 — a destructuring pattern nests its parens inside a
  `multi_variable_declaration`, so no paren is ever a direct child of
  `lambda_parameters`.
- **`is_bare_param`** is the one path the new predicate never runs on.
  `bca dump` shows why it is safe: in `x /* c */ -> x` the
  `block_comment` is a *sibling* of the bare `identifier`. The comment is
  therefore not a discriminating input, so that test is written as a
  parity assertion between the commented and bare spellings and says so
  in its doc comment rather than implying coverage it does not have.
- **Corpus.** 854 recorded values move, every one downward, no other
  metric touched. Hand-checked against source: kenlm's
  `DontBhiksha::DontBhiksha` has 4 real parameters and 3 `/*…*/`
  comments (7 → 4), `ReadBackoff` has 2 and 1 (3 → 2).
- `make pre-commit` passes. Patch coverage on the changed production
  lines is 100%, against a project line coverage of 94.04%.

## Notes for review

- **Metric drift**, called out in the CHANGELOG: serialized `nargs`
  falls wherever a signature carries a comment.
- **`.bca-baseline.toml` ratchets *down*.** Folding the four counting
  loops into one dropped `src/metrics/nargs.rs` from 547 to 536
  `loc.ploc`. The refresh also drops four entries left stale by #1203.
- **Snapshot submodule** `dekobon/big-code-analysis-output` is pushed
  (`89e4c973`) and its SHA recorded in this commit.
- **#1200** moves `nargs` on the same C/C++ corpus in the opposite
  direction. It is deliberately not in this branch; landing it after
  this will need one snapshot re-refresh.

### DIFF
--- a/.bca-baseline.toml
+++ b/.bca-baseline.toml
@@ -1029,37 +1029,13 @@ value = 5.0
 path = "src/metrics/nargs.rs"
 qualified = "<file>"
 metric = "loc.ploc"
-value = 547.0
-
-[[entry]]
-path = "src/metrics/npa/objc.rs"
-qualified = "ObjcCode::compute"
-metric = "cognitive"
-value = 15.0
-
-[[entry]]
-path = "src/metrics/npa/php.rs"
-qualified = "PhpCode::compute"
-metric = "halstead.effort"
-value = 51274.96613103099
+value = 536.0
 
 [[entry]]
 path = "src/metrics/npa/python.rs"
 qualified = "python_self_attr_name_bytes"
 metric = "nexits"
 value = 6.0
-
-[[entry]]
-path = "src/metrics/npm/go.rs"
-qualified = "GoCode::compute"
-metric = "halstead.effort"
-value = 65898.2477704053
-
-[[entry]]
-path = "src/metrics/npm/objc.rs"
-qualified = "ObjcCode::compute"
-metric = "cognitive"
-value = 15.0
 
 [[entry]]
 path = "src/node.rs"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -766,6 +766,25 @@ for historical reference.
 
 ### Fixed
 
+- A comment written inside a parameter list is no longer counted as a
+  parameter (#1201). tree-sitter attaches such a comment as a direct
+  child of the parameter-*list* node rather than inside the parameter it
+  documents, and every `nargs` filter listed punctuation only, so
+  `int h(int a /* one */, int b /* two */)` reported 4 and the C++ idiom
+  for a deliberately unused parameter, `void f(int /*unused*/)`,
+  reported 2. Fixed for C, C++, Mozcpp, Objective-C, JavaScript, MozJS,
+  TypeScript, TSX, Python, Rust, Java, C#, PHP, Ruby, Groovy, Elixir and
+  Kotlin lambdas in one shared predicate. Go, Lua, Tcl, iRules,
+  Objective-C methods and blocks, Kotlin functions and Groovy closures
+  were already correct and are unchanged; Perl had carried the exclusion
+  privately since its signature support landed.
+
+  **Metric drift.** Serialized `nargs` falls wherever a signature
+  carries a comment — across the `DeepSpeech` corpus, 854 recorded
+  values, including kenlm's `DontBhiksha::DontBhiksha` (7 → 4) and
+  `ReadBackoff` (3 → 2). A signature that previously tripped
+  `bca check --threshold nargs=N` on its comments alone now passes.
+
 - **Serialized shape.** `npm` and `npa` emission is now decided by the
   space's kind alone, for every language (#1203). #1197 declared the rule
   — containers and the file `unit` root carry the block, a function space
--- a/big-code-analysis-book/src/metrics.md
+++ b/big-code-analysis-book/src/metrics.md
@@ -742,7 +742,9 @@ comparable numbers. The serialised output
 `closure_args_max`.
 The implementation handles default arguments, variadic arguments,
 keyword-only arguments, and destructured parameters consistently per
-language.
+language. Comments written inside the parameter list are not
+parameters, including the C++ idiom that puts one where an unused
+parameter's name would go — `void f(int /*unused*/)` is one argument.
 
 ### What the threshold gate measures {#nargs-gate}
 
--- a/src/metrics/nargs.rs
+++ b/src/metrics/nargs.rs
@@ -208,22 +208,43 @@ impl Stats {
     }
 }
 
+/// How many of `params`' children are formal parameters.
+///
+/// The one place that says what a parameter is, and the only caller of
+/// [`Checker::is_non_arg`]. No language's `is_non_arg` lists a comment —
+/// they carry punctuation plus, in a few cases, a non-parameter the
+/// grammar puts in the list anyway (Rust's `self` receivers, Python's
+/// PEP 570 `/` marker, PHP's `...`) — so the purely negative filter
+/// this replaces counted a comment sitting between two parameters —
+/// `int h(int a /* one */, int b)` reported 3 — and counted the comment
+/// that stands in for an unnamed parameter, so `void f(int /*unused*/)`
+/// reported 2 (#1201). tree-sitter attaches a comment as a direct child
+/// of the parameter list, not inside the parameter it documents, which
+/// is why no `is_non_arg` list could have caught it.
+///
+/// Excluding comments here rather than in twenty `is_non_arg` impls is
+/// what makes it one rule: the next language added inherits it. Perl,
+/// Elixir and Kotlin lambdas reach their parameter list by three routes
+/// `compute_args` cannot express, so they call this directly.
+#[inline]
+fn count_args<T: Checker>(params: &Node) -> usize {
+    params
+        .children()
+        .filter(|child| !T::is_non_arg(child) && !T::is_comment(child))
+        .count()
+}
+
 #[inline]
 fn compute_args<T: Checker>(node: &Node, nargs: &mut usize) {
     if let Some(params) = node.child_by_field_name("parameters") {
         // The field can hold a lone parameter rather than a list, in
-        // which case there are no children to walk and the loop below
+        // which case there are no children to walk and `count_args`
         // yields zero — see `Checker::is_bare_param` (#1185).
         if T::is_bare_param(&params) {
             *nargs += 1;
             return;
         }
-        let node_params = params;
-        node_params.act_on_child(&mut |n| {
-            if !T::is_non_arg(n) {
-                *nargs += 1;
-            }
-        });
+        *nargs += count_args::<T>(&params);
     } else if node.child_by_field_name("parameter").is_some() {
         // JS/TS/TSX/MozJS arrow functions with a bare identifier parameter
         // (`x => …`) use the singular `parameter` field instead of the plural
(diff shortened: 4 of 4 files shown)


dekobon:
## Coverage follow-up (`d5cd307`)

Codecov flagged patch coverage at 92.31% against a project 98.26%. I
reproduced it locally with the CI command
(`cargo llvm-cov --no-report nextest --all-features --workspace --locked`)
and split the patch by module:

| | instrumented | missing | patch % |
|---|---|---|---|
| production | 20 | 0 | **100%** |
| test module | 228 | 13 | 94.3% |

None of the shortfall was production code. All 13 were lines that run
only when the test *fails* — the `format!` inside the failure branch,
the `assert!` arguments (evaluated at panic time), and the
`if !lang.is_enabled() { continue; }` guards' untaken arm.

Three changes, each of which also reads better than what it replaced:
the failure branch now pushes the raw `(lang, source, expected, got)`
tuple with the report built once by `{failures:#?}`; the assertion's
computed arguments are bound eagerly and interpolated by name; the two
`if/continue` guards became an iterator `filter`. The "collect every
failure rather than stopping at the alphabetically first" property is
preserved — reverting the fix still reports `37/43` naming every
language.

**Result: patch 92.31% → 97.81%, and the project dip is gone** (`Misses
-2`, previously `+9`).

### The 4 that remain are not reachable by any test

Three are **comment and doc-comment lines**. `cargo-llvm-cov`'s codecov
output assigns a region value to them because a multi-line region
contributes to every line it spans — e.g. line 4205 is a bare `///`
scoring `0/1`. No test can execute a comment.

The fourth is `failures.push(...)` itself, which no passing run can
reach by construction.

### Worth knowing for future PRs

`codecov.yml`'s `ignore` list has `**/tests/**`, but this repo keeps
most of its tests in `#[cfg(test)]` modules *inside* `src/`, which no
path-based ignore can exclude. So a test-heavy patch is always measured
on its own test code, including that code's failure paths. Nothing to
change here — just the reason patch coverage reads below project
coverage on PRs like this one.

Issue #1209, opened 2026-08-05
fix(c): K&R function with a pointer return type opens no function space (upstream grammar)


dekobon:
## Summary

A K&R-style (pre-ANSI) C function definition whose return type is a
pointer opens **no function space at all** — no `nom` count, no
`nargs`, no `cyclomatic`, nothing. The same function with a plain
return type is analysed normally.

```c
int  krplain(a, b) int a; int b; { return 0; }   /* space `krplain`, nargs 2 */
int *krptr(a, b)   int a; int b; { return 0; }   /* no space at all */
```

## Root cause: upstream grammar, not our wrapper

`tree-sitter-c` 0.24.2's `_old_style_function_definition` is
`_declaration_specifiers` + `field(declarator,
alias(_old_style_function_declarator, function_declarator))` +
`repeat1(declaration)` + body, and `_old_style_function_declarator` is
`_declarator` + `_old_style_parameter_list`. The declarator chain sits
*inside* the old-style declarator, so an outer `pointer_declarator` is
unreachable and GLR prefers the plain `declaration` reading.

**The trigger is any wrapping declarator, not specifically a pointer.**
`int *`, `char **`, `struct S *` and `static int *` all reproduce it.
An ANSI parameter list with a pointer return (`int *f(void) { … }`)
does not. What breaks the parse is a K&R parameter list appearing after
a declarator the return type wrapped.

**There is no `ERROR` node — the parse silently succeeds wrongly, and
the tree is re-associated rather than merely reclassified.** For
`int *krptr(a, b) int a; int b; { … }`:

```
translation_unit
├─ declaration                    "int *krptr(a, b) int a;"
│  ├─ primitive_type: int
│  └─ pointer_declarator
│     └─ function_declarator      absorbs the FIRST K&R decl
│        ├─ identifier: krptr
│        ├─ parameter_list: (a, b)
│        ├─ identifier: int       ← stray
│        └─ identifier: a         ← stray
├─ declaration                    "int b;"  (now top-level)
└─ compound_statement             ← body ORPHANED as a sibling
```

So there is no node for `Checker::is_func` to match — but not because a
`declaration` arm is missing. `src/checker/c.rs:23-36` already
enumerates both `function_definition` aliases (`C::FunctionDefinition |
C::FunctionDefinition2`, the latter being the `_old_style_function_
definition` alias), as do `src/getter/c.rs:19` and `:46`. Nothing in
[...]

Issue #1208, opened 2026-08-04
fix(getter): C-family function returning a function pointer resolves its space name to null


dekobon:
## Summary

A C/C++/Objective-C function whose declared name sits inside a
`parenthesized_declarator` — the shape a function returning a *function
pointer* takes — resolves its `FuncSpace::name` to `null`.

```c
int (*fp(int a, int b))(int c) { return 0; }
```

```json
{ "name": null, "kind": "function", "start_line": 1, ... }
```

A plain `int plain(int a)` reports `"plain"`, and a pointer- or
reference-returning function reports its name correctly too — only the
parenthesised shape fails.

## Root cause

`get_func_space_name` in `src/getter/{c,cpp,mozcpp,objc}.rs` takes the
`declarator` field and then `first_occurrence(FunctionDeclarator*)` —
a leftmost descendant search. For the nested shape that search returns
the **outer** `function_declarator`, the one belonging to the return
type, whose `child(0)` is a `parenthesized_declarator` rather than an
identifier. None of the identifier kinds match, so the arm falls through
and the function returns `None`.

```
function_definition
└─ function_declarator            ← first_occurrence stops here
   ├─ parenthesized_declarator    ← child(0): not an identifier kind
   │  └─ pointer_declarator
   │     └─ function_declarator   ← the one that owns `fp`
   │        ├─ identifier: fp
   │        └─ parameter_list: (int a, int b)
   └─ parameter_list: (int c)
```

## Suggested fix

`declarator_params_owner` in `src/metrics/nargs.rs` (added by #1200)
already walks this chain correctly and returns the innermost declarator
carrying a `parameters` field — for the tree above, the one whose
`child(0)` *is* `fp`. Lifting it somewhere both modules can reach and
using it in place of `first_occurrence` would fix all four getters at
once and drop their `FunctionDeclarator`/`2`/`3` alias enumerations,
which is a `grammar-dispatch.md` §1 improvement in its own right.

Check `get_space_kind` and `is_func_space` for the same shape before
closing — per `.claude/rules/grammar-dispatch.md` §6 the space tree must
not disagree with itself.

## Scope

All four C-family getters share the pattern verbatim, so fix them
together (`grammar-dispatch.md`, "when you fix one language, sweep the
rest").

## Regression test

The nested shape in C, C++, Mozcpp and Objective-C, asserting the space
name is `fp` and not `null`. `src/metrics/nargs.rs`'s
[...]

Issue #1162, opened 2026-08-01
fix(functions): bca functions and find --type function miss source-aware function spaces


dekobon:
## Summary

`#1130` fixed `ops_inner` (`src/ops.rs`), which opened a function space
on the byte-less `is_func || is_func_space` instead of the source-aware
`promotes_to_func_space_with_code` that `spaces::compute::metrics_inner`
uses. Two more seams carry the same bug and were left alone by that fix.

Elixir's `defmodule` / `def` / `defp` / `defmacro` / `defmacrop` are not
distinct grammar productions — they are `Call` nodes whose target
identifier text spells the keyword (#275, #310). A byte-less predicate
cannot see them, so every seam that reaches `Checker::is_func` without
the source bytes reports zero functions for Elixir.

## Affected seams

| Seam | Site | Surface |
|---|---|---|
| `function()` | `src/function.rs:60` | `bca functions`, `Ast::functions` |
| `--type function` filter | `src/parser.rs:191` | `bca find --type function` |

## Reproducer

```elixir
defmodule Foo do
  def bar(x) do
    x + 1
  end
end
```

```console
$ bca metrics --no-config -O json -p t.ex | jq -c '[.. | select(type=="object" and has("kind")) | {kind,name}]'
[{"kind":"unit","name":"t.ex"},{"kind":"class","name":"Foo"},{"kind":"function","name":"bar"}]

$ bca ops --no-config -O json -p t.ex | jq -c '[.. | select(type=="object" and has("kind")) | {kind,name}]'
[{"kind":"unit","name":"t.ex"},{"kind":"class","name":"Foo"},{"kind":"function","name":"bar"}]   # fixed by #1130

$ bca functions --no-config -p t.ex
                                    # <- nothing; exits 0

$ bca find --no-config --type function -p t.ex
                                    # <- nothing; exits 0
```

Both commands work for the equivalent Rust input, so this reads as "this
Elixir file has no functions" rather than as an error.

## Fix sketch

**`src/function.rs:60` is a one-line change.** `function()` already holds
both `code` and `ancestors`:

```rust
if T::Checker::is_func_with_code(n, code, ancestors) {
```

Note `is_func_with_code`, not `promotes_to_func_space_with_code` — this
seam wants *functions*, not every node that opens a space, so a
`defmodule` must stay out (matching the current behaviour for Rust
`impl` blocks and Java classes).

**`src/parser.rs:191` needs a wider change.** The filter is built as a
[...]

Issue #1142, opened 2026-07-31
fix(metrics/nargs): Elixir reports 0 arguments for every function


dekobon:
## Summary

`nargs` reports 0 for every Elixir function, including ones with an
explicit formal parameter list.

```console
$ cat /tmp/t.ex
defmodule Foo do
  def bar(a, b, c) do
    a + b + c
  end
end

$ bca metrics --no-config -O json /tmp/t.ex | jq '.. | .nargs? // empty | .total'
0
0
0
```

Lua, whose function syntax is comparably simple, reports 3 for the
equivalent input, so this is not a general limitation of the default
implementation.

## Why it is currently invisible

The behaviour is anchored by `elixir_default_nargs_is_zero` in
`src/metrics/nargs.rs`, which documents the cause:

> `def` calls are not recognised as functions (`is_func` returns
> `false`), and Elixir anonymous functions hold their parameters inside
> a `stab_clause` (not a parameter-list field), so the default
> `NArgs::compute` heuristic finds no formal parameters to count.

That test exists precisely so a future implementation cannot silently
regress to zero, which is the right call — but the limitation is not
surfaced anywhere a *user* would look. `bca list-metrics` advertises
`nargs`, the book's metric table lists it without qualification, and
`bca check --threshold nargs=5` silently passes every Elixir file.

Found while deriving per-language threshold recommendations for #1140:
Elixir, Bash, and Perl were the three languages whose entire `nargs`
distribution was 0. For Bash and Perl that is correct — neither has
formal parameter lists — but Elixir does.

## Scope

Two separable pieces; the second is worth doing even if the first is
declined.

1. **Implement it.** Count parameters for `def` / `defp` (and the
   `stab_clause` form for anonymous functions). Needs the Elixir
   `is_func` predicate to recognise `def` calls first, which may have
   knock-on effects on `nom` and on which spaces are created — worth
   checking against the existing Elixir metric snapshots before
   committing to the approach.

2. **Document the gap.** Whatever happens to (1), the supported-metrics
   table in `big-code-analysis-book/src/metrics.md` should mark `nargs`
   as unsupported for Elixir, Bash, and Perl, with a one-line reason for
   each — a metric that silently reports 0 reads as "no offenders", not
   as "not measured".

---

### Correction to the scope note above

Scope item 1 says the fix "needs the Elixir `is_func` predicate to recognise
`def` calls first, which may have knock-on effects on `nom` and on which
[...]

PR #1211, opened 2026-08-05
fix(nargs): count args on the innermost C declarator

### LINKED ISSUE #1200: fix(nargs): C/C++ pointer/reference-return functions report nargs = 0

### LINKED ISSUE #1210: fix(nargs): C's (void) marker is counted as a parameter
Fixed on `fix/1200-c-family-nargs-declarator` alongside #1200, in its own commit (`7253f144`).

`int f(void)` declares no parameters, but the grammar emits a real `parameter_declaration` for the `void` and every `nargs` negative filter counted it, so `f(void)` and `f(int)` both reported 1.

The distinction is not structural — an unnamed parameter is the *same* shape (a bare type, no declarator) and really is one argument — so only the bytes separate them, per `.claude/rules/grammar-dispatch.md` §10. `count_args` now takes the source bytes and consults a new `Checker::is_empty_param_marker`, defaulting to `false` and overridden by C, C++, Mozcpp and Objective-C through one shared `c_family_void_parameter` helper. `void *p` keeps counting: it carries a declarator.

Fixtures for all four grammars, with both discriminating controls (`int unnamed(int)` → 1, `int ptr(void *p, int a)` → 2). Perturbing the exclusion away fails the `(void)` rows in all four languages; perturbing the byte comparison to a structural check fails the `unnamed` rows instead.

**Drift:** 24 recorded values fall to 0 across the `DeepSpeech` corpus, all in `pywrapfst.cc` — the only file with `(void)` definitions (`PyThread_tss_alloc(void)` and seven others). Nothing rises. Submodule accepted, committed and pushed (`1b032d7c`).

Found as a pre-existing issue while cross-checking paired plain/pointer probes during #1200.

dekobon:
Fixes #1200
Fixes #1210

C, C++, Mozcpp and Objective-C functions whose return type is a pointer or a reference reported `nargs.function_args = 0` regardless of their real arity — 7.9% of the matched functions in the DeepSpeech/kenlm C++ corpus. Since #1196 made the `nargs` gate read a callable's own parameter count, every one of them was also invisible to `bca check --threshold nargs=N`.

## The issue's prescribed fix would have shipped half the bug

Both the issue body and its resolution plan specify walking the `declarator` **field** down to a node carrying `parameters`. Read against the pinned grammars' `node-types.json`, that repairs the pointer cases and **leaves every reference case at 0** — one of the shapes the issue itself lists and demands fixtures for:

| rule | `declarator` field |
| --- | --- |
| `pointer_declarator` | required |
| `function_declarator` | required |
| `abstract_function_declarator` | **optional** |
| `reference_declarator` | **absent — no fields at all** |
| `parenthesized_declarator` | **absent — no fields at all** |
| `attributed_declarator` | **absent — no fields at all** |

In `reference_declarator` and `parenthesized_declarator` the inner declarator is a plain named child with no field name, so a field-only descent stops dead.

## The rule that shipped

Walk the declarator chain and take the **innermost** node carrying a `parameters` field:

```
step(n):
    n.child_by_field_name("declarator")   -> that child
    else if n has a "parameters" field    -> None   (chain ends)
    else                                  -> last named non-comment non-attribute child
```

…cut at any `operator_cast`, and seeded from the `declarator` field rather than the function node.

Every clause is load-bearing:

- **Field first**, per grammar-dispatch §3, which also sidesteps §1 — `PointerDeclarator2`, `FunctionDeclarator2`/`3`, `ReferenceDeclarator2`/`3`/`4` are aliases a `kind_id` match would have to enumerate.
- **Stop on a `parameters`-bearing node with no `declarator` field** — the C++ lambda. `abstract_function_declarator.declarator` is *optional*, so `[](int a, int (*cb)(int x))` would otherwise return `cb`'s `(int x)`: 1 instead of 2.
- **Last, not sole or first, named child** — `int (__cdecl *w(int a, int b))(int c)` puts a real `ms_call_modifier` ahead of the declarator.
- **Excluding `attribute_declaration`** — `attributed_declarator` is the one fieldless rule putting its declarator *first*, so without this `int f(int a, int b) [[deprecated]]` reports 0. The GNU `__attribute__((…))` spelling was never affected: all four grammars absorb it into the `function_declarator` instead of wrapping it.
- **Cutting at `operator_cast`** — its `declarator` field is the type the operator converts *to*. A conversion operator takes no arguments however many its target type has.

**Innermost, not first**, is what also fixes a function returning a function pointer. `int (*fp(int a, int b))(int c)` had reported 1 — the return type's list — where `fp` takes 2. The issue's plan proposed a `FIXME` for this; the rule gets it for free.

## Consolidated rather than patched in place

`CppCode`, `CCode` and `MozcppCode` carried three byte-identical `NArgs::compute` overrides — which is why one bug existed in triplicate. All three are gone, replaced by a one-line `params_owner`, the hook whose own doc comment says re-stating `compute` is the drift #1142 and #1162 were filed about. The default `compute` now routes closures through it too (a no-op elsewhere: only `JavaCode` overrides it, and only for record compact constructors). This also removed `CCode`'s dead closure branch — `CCode::is_closure` is a constant `false`.

`ObjcCode` keeps its own `compute` (three unrelated parameter shapes) with only the `FunctionDefinition` arm rerouted.

## Also fixed: C's `(void)` marker (#1210)

`int f(void)` declares no parameters, but the grammar emits a real `parameter_declaration` for the `void` and every filter counted it, so `f(void)` and `f(int)` both reported 1. The distinction needs the source bytes — an unnamed parameter is the same shape and really *is* one argument — so `count_args` now threads `&[u8]` to a new `Checker::is_empty_param_marker`, defaulting to `false` and overridden by the four C-family grammars through one shared helper. `void *p` keeps counting: it carries a declarator.

This is an independent behaviour change in its own commit (`7253f144`), reviewable and revertable apart from the rest.

## Verification

Seven perturbations, each applied to one clause alone and run against the whole lib suite, each producing a disjoint failure set:

| perturbation | failures |
| --- | --- |
| field-only descent (the plan's fix) | 22/30 — every pointer, reference and nested row |
| first instead of innermost | the 4 nested `fp` rows |
| drop the lambda stop clause | the 2 lambda-guard rows |
| first instead of last named child | the 4 `__cdecl` rows |
| start off the declarator chain | the 2 local-declaration rows |
| drop the `operator_cast` cut | the 4 conversion-operator rows |
| drop the `attribute_declaration` exclusion | the 3 `[[deprecated]]` rows |
| drop the `(void)` exclusion | the 4 `(void)` rows |
| structural-only `(void)` check | the 4 `unnamed` rows instead |

Two of those fixtures only became real after the *selector* was fixed, and both times the selector was the weak point rather than the assertion:

- The parameterless-lambda row I added to close a coverage gap gave 100% region coverage of the walk's entry guard and failed **0 of 3,322 tests** when that guard was perturbed. Two further candidate fixtures also failed to fail — the `declarator` fields keep pulling the walk back onto declarator-shaped nodes, which is precisely why the guard is safe. The row that discriminates puts a local function *declaration* last in the lambda body.
- Both `operator_cast` fixtures passed with the regression reinstated, because `sole_space_args` read the *first* nested space and a conversion operator sits two levels down, behind its `struct`'s container space. The helper now descends to the innermost sole space.

Also probed as paired plain/pointer variants: constructors, destructors, `operator=`, templates, `[[nodiscard]]`, trailing return types, variadics, `const`-qualified pointer returns, 2000-level pointer nesting (0.00 s — the walk is an iterative `successors` unfold, not recursion), and malformed input.

## Drift

- **#1200**: 3,336 `nargs` values rise across 276 `DeepSpeech` snapshot files. Nothing falls; no other metric key moves.
- **#1210**: 24 values fall to 0, all in `pywrapfst.cc`, the corpus's only `(void)` definitions.

Submodule accepted, committed and pushed (`1b032d7c`), with the SHA recorded in the parent commits alongside each fix.

## Follow-ups filed

- **#1208** — the same nested shape resolves its space name to `null`. A getter bug in a different metric surface; pinned in CI here by `a_parenthesised_declarator_still_loses_its_space_name`, which fails the day it is fixed.
- **#1209** — a K&R function with a pointer return opens no function space at all. Upstream: `tree-sitter-c` parses it as a `declaration`, so there is no `function_definition` for `is_func` to match and nothing in our wrapper to fix.

## Gates

`make pre-commit` → `BCA_GATE: pass`. `make bench-scaling` → all 26 probes within their complexity bound. No public-API change (`NArgs` is `pub(crate)` + `#[doc(hidden)]`), so no `STABILITY.md` cross-reference is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -778,11 +778,35 @@ for historical reference.
   is fixed in the same walk: it had been reporting `(int c)`, the
   return type's list, rather than its own.
 
+  A C++11 `[[…]]` attribute on a function no longer hides its
+  parameters either. `attributed_declarator` is the one declarator rule
+  that puts its declarator *first*, so
+  `int f(int a, int b) [[deprecated]]` had always reported 0. The GNU
+  `__attribute__((…))` spelling was never affected — every one of the
+  four grammars absorbs it into the `function_declarator` instead of
+  wrapping it.
+
+  A C++ conversion operator is explicitly *excluded* from the walk:
+  `operator int (*)(int x)` takes no arguments however many its target
+  type has.
+
   **Metric drift.** Serialized `nargs` rises wherever such a function
   appears — 3,336 recorded values across 276 files of the `DeepSpeech`
   corpus, and nothing falls. Since #1196 made the gate read a
   callable's own parameter count, these functions were invisible to
   `bca check --threshold nargs=N` and can now trip it.
+
+- C's `(void)` marker is no longer counted as a parameter. `int f(void)`
+  declares nothing, but the grammar emits a real `parameter_declaration`
+  for the `void` and every `nargs` filter counted it, so `f(void)` and
+  `f(int)` both reported 1. Fixed for C, C++, Mozcpp and Objective-C.
+  The distinction needs the source bytes rather than the tree — an
+  unnamed parameter is the same shape and really is one argument — so
+  `Checker` gained an `is_empty_param_marker` hook that defaults to
+  `false` and reads them.
+
+  **Metric drift.** 24 recorded values fall to 0 in the `DeepSpeech`
+  corpus, all in `pywrapfst.cc`, its only `(void)` definitions.
 
 - A comment written inside a parameter list is no longer counted as a
   parameter (#1201). tree-sitter attaches such a comment as a direct
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -766,6 +766,24 @@ for historical reference.
 
 ### Fixed
 
+- C, C++, Mozcpp and Objective-C functions whose return type is a
+  pointer or a reference now report their real arity instead of 0
+  (#1200). C declarator syntax nests outward from the declared name, so
+  `FILE *f(int a, int b, int c)` puts the parameter list on a
+  `function_declarator` wrapped by the `pointer_declarator` the return
+  type contributed; `nargs` read the wrapper, found no parameters and
+  reported nothing. `int **`, `int &`, `Foo &&`, `static int *` and a
+  member function returning a reference were all affected. A function
+  returning a *function pointer* — `int (*fp(int a, int b))(int c)` —
+  is fixed in the same walk: it had been reporting `(int c)`, the
+  return type's list, rather than its own.
+
+  **Metric drift.** Serialized `nargs` rises wherever such a function
+  appears — 3,336 recorded values across 276 files of the `DeepSpeech`
+  corpus, and nothing falls. Since #1196 made the gate read a
+  callable's own parameter count, these functions were invisible to
+  `bca check --threshold nargs=N` and can now trip it.
+
 - A comment written inside a parameter list is no longer counted as a
   parameter (#1201). tree-sitter attaches such a comment as a direct
   child of the parameter-*list* node rather than inside the parameter it
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -362,6 +362,24 @@ pub(crate) trait Checker {
     fn is_non_arg(_: &Node) -> bool {
         false
     }
+    /// Whether `param` is a language's way of spelling *"this function
+    /// takes nothing"* inside an otherwise ordinary parameter list.
+    ///
+    /// C's `int f(void)` declares zero parameters, but the grammar
+    /// emits a real `parameter_declaration` for the `void`, so the
+    /// negative filters in [`crate::nargs`] counted it as one — `f` and
+    /// `int f(int)` both reported 1 despite one of them taking no
+    /// argument at all.
+    ///
+    /// It takes `code` because the question is not structural: a
+    /// `parameter_declaration` holding a bare `primitive_type` and no
+    /// declarator is *also* how an unnamed parameter is spelled, and
+    /// `void f(int)` really does take one. Only the bytes tell them
+    /// apart — `.claude/rules/grammar-dispatch.md` §10.
+    #[inline]
+    fn is_empty_param_marker(_param: &Node, _code: &[u8]) -> bool {
+        false
+    }
     /// Whether `node` — the value of a function's `parameters` field —
     /// is itself a *single formal parameter* rather than a parameter
     /// *list*.
@@ -546,6 +564,25 @@ pub(crate) fn csharp_accessor_count(node: &Node) -> usize {
 /// accessor-less expression-bodied form (#464 indexer, #472 property).
 pub(crate) fn csharp_member_has_accessors(node: &Node) -> bool {
     csharp_accessor_count(node) > 0
+}
+
+/// Whether `param` is C's `(void)` marker — the spelling for an empty
+/// parameter list — rather than a parameter.
+///
+/// Shared by C, C++, Mozcpp and Objective-C, which all inherit the
+/// idiom and all emit a real `parameter_declaration` for it.
+///
+/// The declarator check is what separates the marker from `void *p`,
+/// which is an ordinary parameter of type `void *`, and the byte
+/// comparison is what separates it from an unnamed parameter of any
+/// other type: `f(int)` and `f(void)` are the same shape and different
+/// arities.
+pub(crate) fn c_family_void_parameter(param: &Node, code: &[u8]) -> bool {
+    param.child_by_field_name("declarator").is_none()
+        && param
+            .child_by_field_name("type")
+            .and_then(|ty| code.get(ty.start_byte()..ty.end_byte()))
+            == Some(b"void".as_slice())
 }
 
 /// Strip the `#` / `#!` marker plus the `[...]` brackets from a
--- a/src/checker/c.rs
+++ b/src/checker/c.rs
@@ -44,6 +44,11 @@ impl Checker for CCode {
         node.kind_id() == C::CallExpression
     }
 
+    // C's `(void)` marker, shared with C++, Mozcpp and Objective-C.
+    fn is_empty_param_marker(param: &Node, code: &[u8]) -> bool {
+        c_family_void_parameter(param, code)
+    }
+
     fn is_non_arg(node: &Node) -> bool {
         matches!(node.kind_id().into(), C::LPAREN | C::COMMA | C::RPAREN)
     }
--- a/src/checker/cpp.rs
+++ b/src/checker/cpp.rs
@@ -55,6 +55,11 @@ impl Checker for CppCode {
         node.kind_id() == Cpp::CallExpression
     }
 
+    // C's `(void)` marker, shared with C++, Mozcpp and Objective-C.
+    fn is_empty_param_marker(param: &Node, code: &[u8]) -> bool {
+        c_family_void_parameter(param, code)
+    }
+
     fn is_non_arg(node: &Node) -> bool {
         matches!(
             node.kind_id().into(),
--- a/src/checker/mozcpp.rs
+++ b/src/checker/mozcpp.rs
@@ -55,6 +55,11 @@ impl Checker for MozcppCode {
         node.kind_id() == Mozcpp::CallExpression
     }
 
+    // C's `(void)` marker, shared with C++, Mozcpp and Objective-C.
+    fn is_empty_param_marker(param: &Node, code: &[u8]) -> bool {
+        c_family_void_parameter(param, code)
+    }
+
     fn is_non_arg(node: &Node) -> bool {
         matches!(
             node.kind_id().into(),
--- a/src/checker/objc.rs
+++ b/src/checker/objc.rs
@@ -60,6 +60,11 @@ impl Checker for ObjcCode {
         )
     }
 
+    // C's `(void)` marker, shared with C++, Mozcpp and Objective-C.
+    fn is_empty_param_marker(param: &Node, code: &[u8]) -> bool {
+        c_family_void_parameter(param, code)
+    }
+
     fn is_non_arg(node: &Node) -> bool {
         matches!(
             node.kind_id().into(),
--- a/src/metrics/nargs.rs
+++ b/src/metrics/nargs.rs
@@ -4629,16 +4629,12 @@ mod c_family_return_type_declarators {
         // fixtures assert about the struct and pass with the defect
         // reinstated.
         let mut space = &root;
-        loop {
-            let [only] = space.spaces.as_slice() else {
-                break;
-            };
+        let mut depth = 0;
+        while let [only] = space.spaces.as_slice() {
             space = only;
+            depth += 1;
         }
-        assert!(
-            !std::ptr::eq(space, &root),
-            "{lang:?}: fixture opened no space at all"
-        );
+        assert!(depth > 0, "{lang:?}: fixture opened no space at all");
         (
             space.metrics.nargs.closure_args(),
             space.metrics.nargs.function_args(),
--- a/src/metrics/nargs.rs
+++ b/src/metrics/nargs.rs
@@ -303,28 +303,23 @@ fn compute_args<T: Checker>(node: &Node, nargs: &mut usize) {
 /// Every step strictly descends a finite tree, so the walk terminates
 /// without a depth cap.
 fn declarator_params_owner<'tree, T: Checker>(node: &Node<'tree>) -> Option<Node<'tree>> {
-    // The chain starts at the field rather than at `node` so the
-    // last-named-child fallback can never fire on the function node
-    // itself and walk into its body.
-    let mut current = node.child_by_field_name("declarator")?;
-    let mut owner = None;
-    loop {
-        let carries_params = current.child_by_field_name("parameters").is_some();
-        if carries_params {
-            owner = Some(current);
-        }
-        let next = match current.child_by_field_name("declarator") {
+    // The chain starts at the `declarator` field rather than at `node`
+    // so the last-named-child fallback can never fire on the function
+    // node itself and walk into its body. The walk runs outside-in, so
+    // the innermost qualifying link is the last one it yields.
+    std::iter::successors(
+        node.child_by_field_name("declarator"),
+        |current| match current.child_by_field_name("declarator") {
             Some(declarator) => Some(declarator),
-            None if carries_params => None,
+            None if current.child_by_field_name("parameters").is_some() => None,
             None => current
                 .children()
                 .filter(|child| child.is_named() && !T::is_comment(child))
                 .last(),
-        };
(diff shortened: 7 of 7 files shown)


dekobon:
## Review: PR #1211 (`fix/1200-c-family-nargs-declarator`, 10 commits)

Full review per the `review` skill checklist, plus independent empirical verification of the walk against parse shapes outside the committed fixture tables. **Verdict: APPROVE WITH COMMENTS** — one trivial docs fix, two optional test-gap suggestions, no bugs found.

### Independently verified

- **The walk's correctness beyond the fixtures.** Probed with the branch binary against shapes the tables don't pin, all correct:
  - Out-of-line qualified conversion operator `S::operator int (*)(int x) { … }` → 0. The AST routes it as `function_definition.declarator = qualified_identifier` whose *last named child* is the `operator_cast`, so the `take_while` cut fires via the fallback path, not the seeded field — a different chain than the in-class fixtures exercise.
  - Trailing-return function pointer `auto f(int a, int b) -> int (*)(int c)` → 2 (the stop-on-`parameters` clause keeps the walk out of the trailing return).
  - Lambda with fn-pointer trailing return `[](int a) -> int (*)(int q)` → closure 1.
  - `(void)` lambda `[](void){ … }` → closure 0 — a shape the fix handles *for free* because the closure channel now routes through the same `count_args`/marker path.
  - Redundant parens `int (f)(int a)` → 1; `int *f(int a) noexcept` → 1; comment inside the marker `int f(/* nothing */ void)` → 0; 50-level pointer nesting → instant; truncated `int *f(` → no panic, no space.
- **One perturbation replicated.** Removed the `operator_cast` `take_while` cut and re-ran: exactly **4/61 rows fail** — the two conversion-operator fixtures × {Cpp, Mozcpp} — matching the verification table's claimed failure set. Restored; tree clean; full nargs module (149 tests) green.
- **Submodule drift matches the claims**: `89e4c973..28dd2dec` touches 276 files (#1200); `28dd2dec..1b032d7c` touches only the two `pywrapfst.cc` snapshots (#1210). The recorded SHA is on the submodule's `origin/main`.
- **The `params_owner` doc inventory is exact**: compute-overriders are precisely Objc, Go, Kotlin, Lua, Tcl, iRules, Perl, Elixir, Groovy; the only other `params_owner` override is `JavaCode`, and it is kind-guarded (`CompactConstructorDeclaration`), so routing the closure channel through `params_owner` cannot misfire on Java lambdas.
- **CHANGELOG entries sit inside `[Unreleased]`** (section spans lines 25–1546; the edited `### Fixed` is at 767).
- **`[crate::nargs]` doc link resolves** (the module is re-exported; `parser.rs`/`traits.rs` already use that path). Worth knowing it would *not* have been caught if broken: `Checker` is `pub(crate)`, so the `cargo doc -D warnings` gate never sees its links.
- The vacuity-guard asserts (`shared > 0 && …`) can't fire on the feature-matrix CI legs — those run `cargo check --all-targets` only — and the pattern matches the established #1201 module convention.

### Docs

| # | Finding | File | Effort |
|---|---------|------|--------|
| 1 | The `(void)`-marker CHANGELOG entry cites no issue number. The sibling entry ends its first sentence with “(#1200)”; this one never mentions #1210, though the PR closes it and the commit message references it. | `CHANGELOG.md` (the `### Fixed` entry beginning “C's `(void)` marker…”) | trivial |

### Test gaps (optional; non-blocking)

| # | Finding | File | Effort |
|---|---------|------|--------|
| 2 | No committed fixture pins the **out-of-line/qualified conversion operator**. It works today, but through a *different* chain than the two in-class fixtures (the cut fires off `qualified_identifier`'s last-named-child fallback rather than off the seeded `declarator` field), so a grammar bump that reshapes qualified operator-cast parsing could regress it while both existing `operator_cast` rows stay green. One row in `cpp_only_shapes` covers it: `"struct S { operator int (*)(int x); }; S::operator int (*)(int x) { return nullptr; }"` → `(0, 0)`. | `src/metrics/nargs.rs` (`cpp_only_shapes`) | trivial |
| 3 | Three shapes the PR narrative says were probed but that no committed row pins: trailing-return fn-pointer (`auto f(int a, int b) -> int (*)(int c)` → 2 — exercises the stop clause from the *function* channel, which no current fixture does), the `(void)` lambda (`[](void){}` → closure 0 — the marker on the closure channel), and `int (f)(int a)` → 1. All verified correct on this branch; the tables make adding rows cheap. | `src/metrics/nargs.rs` | trivial |

### Pre-existing observations (not introduced here, no action needed)

- `int v(int a, ...)` reports **2** — the `...` counts as a parameter. Unchanged by this PR (the plain-return path counted it identically before), and the fix keeps plain/pointer variants consistent, but it's a semantic choice nobody has pinned in a test or doc.
- The probes reproduce the known name-resolution gaps the PR already tracks: `int (f)(int a)` and the out-of-line conversion operator both report `name: null` (#1208 family), and K&R with pointer return opens no space (#1209).
- The PR body's perturbation table row counts (e.g. “22/30”) reflect the table size at the commit each perturbation ran against; the final suite has 61 rows. Narrative only.

### Checklist areas with nothing to report

Correctness of the marker predicate (`void *p`, unnamed `int`, `optional_parameter_declaration`, `variadic_parameter` all correctly non-markers; byte read is bounds-safe via `code.get`), termination (iterative `successors`, strictly descending), performance (O(depth + width) per function node, no allocation), security (no panics on malformed input), API surface (`NArgs` is `pub(crate)` + `#[doc(hidden)]`, no `STABILITY.md` impact), and the `code`-threading changes to Kotlin/Perl/Elixir/Groovy are pure plumbing (marker defaults to `false`).

### Summary

- Files reviewed: 8 (all, in full) + submodule diff
- Findings: 1 docs (trivial), 2 test-gap (optional)
- Verdict: **APPROVE WITH COMMENTS**

Issue #1228, opened 2026-08-07
chore(lints): enums is CI-linted without the workspace pedantic set


dekobon:
> **Status: fixed** on `fix/issue-1228` (commit `2ee1946e`), pending merge.
> The 23-warning table below is confirmed for pedantic, but it undercounts:
> `missing_docs` is also in the workspace set and adds 15 more, so the real
> cost was **38**. Both the lints table and the `check-excluded-manifests.py`
> extension shipped; the five vendored `tree-sitter-*` crates were decided
> **exempt**. See the comment below for the full breakdown.

Found while completing #1227's `unwrap` gate, which had the same shape of gap.

## The gap

`[workspace.lints]` in the root `Cargo.toml` gives every workspace member
`pedantic = "warn"` plus `missing_docs = "warn"`. #158 propagated it and was
scoped, in its own words, to "all three shipping crates".

`enums` is not one of them. It is excluded from `[workspace] members` (it is a
separate workspace, so it *cannot* inherit `lints.workspace = true`), and it
carries no `[lints]` table of its own. The five vendored `tree-sitter-*` crates
are in the same position.

This is not a crate nobody lints. `make enums-check` — wired into `make
pre-commit` and `make ci` — runs:

```
RUSTFLAGS="-D warnings" cargo clippy --manifest-path enums/Cargo.toml \
  --all-targets --locked -- -D warnings
```

So `enums` is gated as hard as any shipping crate, just against a **smaller
lint set**. The result reads as "fully linted" and is not.

## Measurement

`cargo clippy --manifest-path enums/Cargo.toml --all-targets -- -W clippy::pedantic`
reports **23** warnings:

| Count | Lint |
|-------|------|
| 5 | `must_use_candidate` |
| 4 | `missing_errors_doc` |
| 7 | `wildcard_imports` |
| 3 | `needless_pass_by_value` |
| 2 | `uninlined_format_args` |
| 1 | `match_same_arms` |
| 1 | `missing_panics_doc` |

23 is small enough to clear in one sitting, unlike the ~2,200 that made #158 a
multi-PR effort for the shipping crates.

## Suggested fix

Add an explicit `[lints.clippy]` / `[lints.rust]` table to `enums/Cargo.toml`
mirroring the workspace set, then triage the 23 with the tier heuristic in
`.claude/skills/rust-optimize` (the `wildcard_imports` hits are prelude-style
`use crate::common::*` and are the likely `allow`).

[...]

Issue #158, opened 2026-05-11
refactor(workspace): propagate pedantic clippy + missing_docs lints (carved out of #150)


dekobon:
Carved out of #150.

## Goal

Propagate the workspace-wide pedantic clippy lint set + `missing_docs`
to all three shipping crates so the workspace enforces the same
lint posture as `host-identity`. The metadata refactor portion of
#150 (workspace.package inheritance, MSRV 1.94, resolver=3, profile
changes, publish=false) has landed; this issue covers the lint
propagation half that #150 originally bundled.

## Why this is split out

When #150's lint propagation was first attempted, enabling
`pedantic = { level = "warn", priority = -1 }` plus
`missing_docs = "warn"` with only the host-identity carve-outs
(`module_name_repetitions = "allow"`,
`missing_errors_doc = "allow"`) produced **1844 clippy errors in
the `big-code-analysis` lib alone** and ~2200 workspace-wide once
the CLI and web crates were added. With
`cargo clippy --workspace --all-targets -- -D warnings` (the
`make pre-commit` clippy gate), every one of those becomes a hard
failure.

This is a multi-PR cleanup that warrants its own branch — bundling
it into #150 either (a) blocks the metadata refactor behind a
hundreds-of-files cleanup or (b) lands the lint config in a state
that breaks every subsequent CI run until the cleanup completes.
Splitting it keeps #150 landable and gives this work the space it
needs.

### Add to root `Cargo.toml`

```toml
[workspace.lints.rust]
missing_docs = "warn"

[workspace.lints.clippy]
pedantic = { level = "warn", priority = -1 }
module_name_repetitions = "allow"
missing_errors_doc = "allow"
```

### Add to each shipping crate's `Cargo.toml`

```toml
[lints]
workspace = true
```

Root `big-code-analysis`, `big-code-analysis-cli`,
`big-code-analysis-web`. **NOT** the 5 vendored grammars or
`enums` — those follow upstream conventions and are out of scope.

### Clean up every warning

Warning histogram captured at the start of the cleanup (workspace
totals via `cargo clippy --workspace --all-targets --all-features
-W clippy::pedantic -A clippy::module_name_repetitions
-A clippy::missing_errors_doc`):

| Count | Lint |
|-------|------|
| 1139 | `clippy::float_cmp` (overwhelmingly `assert_eq!(metric.X, 3.0)` in test modules) |
[...]

Issue #126, opened 2026-05-07
bug(c_macro): DOLLARS buffer overflow panics on macro identifiers longer than 2048 bytes


dekobon:
## Summary

The `replace()` function in `c_macro.rs` uses a fixed-size buffer `DOLLARS: [u8; 2048]` to generate dollar-sign replacements for macro identifiers. If an identifier matching a macro exceeds 2048 bytes, the slice `&DOLLARS[..(i - start)]` indexes out of bounds and panics.

## Location

- `src/c_macro.rs:5` — `DOLLARS` declaration
- `src/c_macro.rs:39` — slice used inside the byte loop
- `src/c_macro.rs:55` — slice used for trailing identifier after the loop

## Evidence

```rust
const DOLLARS: [u8; 2048] = [b'$'; 2048];
// ...
new_code.extend(&DOLLARS[..(i - start)]);   // panics if i - start > 2048
```

Given a source file containing a predefined-macro name (or a user-supplied macro) longer than 2048 ASCII characters, `i - start` exceeds 2048 and the indexing panics. The `code` parameter is ultimately user-supplied (source files), so this is reachable with pathological input.

## Expected Behavior

The function should handle identifiers of arbitrary length without panicking. Options:
1. Use `iter::repeat(b'$').take(len)` instead of a fixed-size array.
2. Truncate or skip replacement for unreasonably long identifiers.

## Actual Behavior

The function panics with `index out of bounds` when processing source files containing macro identifiers longer than 2048 bytes.

## Impact

Any caller passing a C/C++ source file with an extremely long identifier matching a predefined macro (or a user-supplied macro set) triggers an unrecoverable panic. This violates the project convention of no panics in non-test code and is a potential denial-of-service vector for the web service and CLI.

dekobon:
## General Plan for Fix

### Root Cause

`replace()` allocates a fixed-size 2048-byte array of `b'$'` bytes at compile time and then slices it at `[..len]` without bounds-checking. If a matching identifier is longer than 2048 bytes, the slice panics.

### Implementation Steps

1. Remove the `DOLLARS` constant and replace the two `extend(&DOLLARS[..n])` sites with an iterator:
   ```rust
   new_code.extend(std::iter::repeat(b'$').take(len));
   ```
   This produces the same output for any length without panic.
2. Remove `const DOLLARS: [u8; 2048] = [b'$'; 2048];` from the top of the file.
3. No change to the function signature or callers is needed.

[...]

PR #1229, opened 2026-08-07
chore(lints): gate span arithmetic, lexer indexing, and production unwrap

### LINKED ISSUE #1152: chore(lints): enable arithmetic and indexing lints on attacker-controlled paths
Split from #1053 (§1). See that issue for the threat model and the explicit
Scorecard non-goal.

## Summary

The no-`unwrap` / no-`panic!` rule in `AGENTS.md` is convention, not
machine-enforced. `[workspace.lints]` in the root `Cargo.toml` configures only:

```toml
[workspace.lints.rust]
missing_docs = "warn"

[workspace.lints.clippy]
pedantic = { level = "warn", priority = -1 }
module_name_repetitions = "allow"
```

Nothing configures `unwrap_used`, `expect_used`, `indexing_slicing`, or
`arithmetic_side_effects`.

This is the cheapest item in #1053 and the best-evidenced: #1051 (a 5-byte
input panicking the library in release, via a `usize` underflow in
`src/metrics/loc/`) would have been caught **statically**, with no fuzzer and
no runtime cost.

## Why this is now replayable, not hypothetical

#1051 is fixed and on `main`. So the claim that
`clippy::arithmetic_side_effects` catches it can be *verified* rather than
argued: enable the lint, check out the pre-#1051 commit, and confirm it flags
that subtraction. **Do that before adopting the lint.** A lint adopted on the
strength of an argument and then drowned in `#[allow]`s is worse than no lint,
because it reads as coverage.

## Scope

> **Status (2026-08-07): all four items done** on
> `fix/1152-1227-panic-free-lints` (item 3 landed with #1227 on the same
> branch). Full findings in the comments below — including that item 2's
> "unproven" premise was wrong: #126 is a replayable `indexing_slicing`
> catch.

- [x] Enable `clippy::arithmetic_side_effects` — scoped to **`src/metrics/loc/`**,
      not blanket `src/metrics/`. The wider scope gives 313 hits of which 267
      are `+=` on other metrics' counters; all 15 subtractions are in `loc`.
      Triaged to **zero** `#[allow]`s in the span code (36 increments collapsed
      into `Lloc::count_logical_line`, the rest made saturating) plus one
      file-level carve-out in `line_set.rs`, where saturating a word index
      would mask a corrupt read rather than prevent one.
- [x] **Validated against `92680fa8^`**: the lint flags both #1051 panic sites,
      `loc/rust.rs:48` (`end - 1`) and `loc/shared.rs:46` (`end - start`).
- [x] `clippy::indexing_slicing` adopted on `src/c_macro.rs`. **The "no
      historical bug" premise was wrong** — #126 was `&DOLLARS[..(i - start)]`
      panicking on a 2049-byte macro identifier, and replaying at `f59c5c06^`
      flags both slices. Nine per-*function* carve-outs so new functions are
      covered by default; `step_raw_string`'s cross-call delimiter slice is
      hardened with `get` rather than allowed.
- [x] Counted `unwrap_used` / `expect_used`: **52 production sites** (not
      several hundred — that is the naive grep), **37 after this PR**. Small
      enough to drive to zero, but adopting means resolving all 37 at once
      under `-D warnings`. Split to **#1227** with the per-file breakdown.
- [x] **`FEATURES_PINNED` removed — all 16 sites**, not just `dispatch.rs`.
      `big-code-analysis-web/src/web/server/handlers.rs` carried the identical
      latent panic at 7 more. Both crates propagate onto the error channel they
      already had; lessons-learned #26, which recommended this pattern, is
      corrected.

## Acceptance

Each adopted lint has a recorded validation result (which historical bug it
catches, or an explicit note that it catches none and is adopted for
prevention), and the `#[allow]` count is small enough to read in one sitting.

## Assessment

| Dimension  | Rating |
|------------|--------|
| Difficulty | Medium |
| Complexity | Low |
| Priority   | Medium |

**Difficulty — Medium.** Enabling the lints is trivial; triaging the hits
across `src/metrics/` and the byte lexers, and deciding each `#[allow]`, is
not. The `FEATURES_PINNED` fix is separate and small.

**Complexity — Low.** Build configuration plus scattered annotations. No public
API, no metric values, no snapshots.

**Priority — Medium.** This is the cheapest item split out of #1053 and the
only one whose value is demonstrable rather than prospective — it catches the
#1051 class statically, for free, on every build. Not High: the specific bug is
already fixed, so this is prevention.

---

## Resolution Plan

Not started. `fix/batch-2026-07-31-2` leaves `[workspace.lints.clippy]` as
`pedantic` plus one allow, so the gap described above is live.

### Order the four lints by evidence, not by strictness

Each has a different amount of proof behind it, and adopting them together
would bury the strong case in the weak ones:

1. **`clippy::arithmetic_side_effects` on `src/metrics/`** — replayable
   evidence (#1051). Do this first, alone.
2. **`clippy::indexing_slicing` on the byte lexers** — plausible but unproven;
   ~20 raw index operations in `c_macro.rs` on attacker-controlled bytes, none
   of which has produced a bug. Validate before adopting.
3. **`expect_used` / `unwrap_used`** — needs a count before it is even a
   decision.
4. **The `FEATURES_PINNED` fix** — independent of all three; do it regardless.

### Steps

1. **Validate `arithmetic_side_effects` against history first.** Enable it
   scoped to `src/metrics/`, check out the commit *before* #1051's fix, and
   confirm it flags the `usize` subtraction in `src/metrics/loc/`. If it does
   not, the scoping is wrong — widen it or reconsider — rather than adopting a
   lint that catches nothing and then allowing away its false positives.
   Record the result in the PR either way; it is the whole justification.
2. Triage the hits. Every `#[allow]` carries a why-comment naming the bound
   that makes the arithmetic safe. If the allow count is large enough that
   nobody will read them all, that is evidence the scope is too wide — narrow
   to the modules that parse attacker-controlled input rather than blanket
   `src/metrics/`.
3. Repeat (1) and (2) for `indexing_slicing`, validated against `c_macro.rs`.
   There is no historical bug to replay here, so the validation is weaker by
   nature — say so, and consider adopting it at `warn` without a CI gate until
   it has caught something.
4. **Count the production `expect(` / `unwrap(` sites properly** — excluding
   `#[cfg(test)]` modules and `tests/` — before deciding. A naive grep over the
   workspace returns several hundred, almost all in tests; the real number
   decides whether `expect_used` is a `warn` you can drive to zero or a
   permanent allow-list that weakens the signal.
5. **Fix the `FEATURES_PINNED` expects regardless.** The
   `.expect(FEATURES_PINNED)` calls in `big-code-analysis-cli/src/dispatch.rs`
   are sound only while `MetricsError` has no parse-failure variant, and that
   enum is `#[non_exhaustive]` with a doc reserving the right to add variants
   in a minor release. That is a latent panic with a scheduled trigger; it
   should not wait on a lint decision. Replace with a `match` that propagates,
   or document the invariant in the `expect` message per the `AGENTS.md`
   carve-out — the current message does not.
6. Lints go in `[workspace.lints.clippy]` where they apply workspace-wide, or
   as module-level `#![warn(...)]` where scoped. Prefer the module-level form
   for `src/metrics/` so the carve-out stays visible at the affected site,
   matching the comment already in the root `Cargo.toml`: *"the allow set is
   intentionally minimal — additional allows go at file or function level so
   the carve-out stays visible."*
7. `make pre-commit` in full. A new `-D warnings`-eligible lint fires in CI
   before it fires locally if the local run is a subset.

## Assessment

Ratings unchanged from the table above (Medium / Low / Medium).

### LINKED ISSUE #1227: chore(lints): triage the 37 production unwrap/expect sites, then gate them
Split from #1152 (scope item 3), which asked for a proper production count
before `unwrap_used` / `expect_used` was "even a decision". Here is the count
and the decision it points to.

## The count

A naive grep over the workspace returns **723**, almost all in test modules.
The real number is what `cargo clippy --workspace --all-features` reports with
the lints enabled: that compiles lib and bin targets only, so `#[cfg(test)]`
modules are excluded by construction rather than by a heuristic.

```bash
cargo clippy --workspace --all-features --message-format=short \
  -- -W clippy::unwrap_used -W clippy::expect_used
```

**52 production sites before #1152, 37 after** — that PR removed the fifteen
`.expect(FEATURES_PINNED)` calls, which were 29% of the total on their own.

| Count | File |
|-------|------|
| 4 | `src/vcs/classify.rs` |
| 4 | `src/tools.rs` |
| 4 | `src/preproc.rs` |
| 4 | `src/ast.rs` |
| 3 | `src/spaces/compute.rs` |
| 2 each | `src/vcs/options.rs`, `src/vcs/git/identity.rs`, `src/ops.rs`, `src/node/parser_cache.rs`, `src/c_macro.rs` |
| 1 each | `src/vcs/identity.rs`, `src/node.rs`, `src/checker.rs`, `src/checker/python.rs`, `xtask/src/main.rs`, `big-code-analysis-web/src/web/function.rs`, `big-code-analysis-cli/src/commands/scan.rs`, `big-code-analysis-cli/src/dispatch.rs` |

## What the number means

37 is small enough to drive to zero — this is a `warn` you can finish, not a
permanent allow-list. That was the open question, and it is now answered.

It is not, however, a mechanical change. `AGENTS.md` already permits
`expect("reason")` for a provably-unreachable invariant with the invariant in
the message, so each of the 37 needs the same judgement #1152 applied to
`FEATURES_PINNED`: is the invariant real, is it documented, and is there an
error channel to propagate onto instead? #1152 found the answer was "propagate"
for all fifteen of its sites. Some of the remaining 37 will be genuine
carve-outs — `c_macro.rs`'s `.expect("invariant: bytes filtered to ASCII-only
by is_identifier_part")` is already in the documented-invariant form.

## Why it is not in #1152

Adopting the lint means resolving all 37 in one pass, because this workspace
runs `-D warnings` in CI — there is no "adopt at `warn` without a gate"
option, which is what #1152's plan suggested for the weaker-evidenced items.
Bundling that with #1152's two validated lint adoptions would have buried
them.

## Suggested approach

1. Triage the 37 into *propagate* / *documented invariant* / *genuinely
   unreachable* before enabling anything.
2. Fix the propagate set first, per crate, so each PR stays reviewable.
3. Enable `unwrap_used` / `expect_used` in `[workspace.lints.clippy]` only
   once the residue is small and every survivor carries its invariant in the
   `expect` message.

Note `unwrap_used` and `expect_used` should be weighed separately: `unwrap()`
carries no invariant by construction and is the stricter, cheaper win.

## Assessment

| Dimension  | Rating |
|------------|--------|
| Difficulty | Medium |
| Complexity | Low |
| Priority   | Low |

Priority is Low, not Medium: unlike #1152's two lints, neither of these has a
historical bug in this repository to replay. The value is prospective.

dekobon:
Closes #1152. Closes #1227.

Two clippy lints adopted on the paths that handle attacker-controlled input,
each **validated by replaying it against the tree that carried the bug it
claims to catch**, plus removal of the fifteen `.expect(FEATURES_PINNED)` call
sites the CLI and web crates used to discharge a widened `Result`.

## Validation, not argument

#1152 asked for this explicitly: *"A lint adopted on the strength of an
argument and then drowned in `#[allow]`s is worse than no lint, because it
reads as coverage."* Both adoptions replay a real historical bug.

| Lint | Replayed at | Result |
|---|---|---|
| `arithmetic_side_effects` | `92680fa8^` (pre-#1051) | flags **both** panic sites: `loc/rust.rs:48` (`end - 1`) and `loc/shared.rs:46` (`end - start`) |
| `indexing_slicing` | `f59c5c06^` (pre-#126) | flags **both** `&DOLLARS[..(i - start)]` slices |
| `unwrap_used` | n/a — 0 production sites | free to adopt; fails CI on the first one added |

**The issue was wrong about the second one.** #1152 rated `indexing_slicing`
"plausible but unproven; ~20 raw indexes in `c_macro.rs`, **none of which has
produced a bug**." But #126 — *"DOLLARS buffer overflow panics on macro
identifiers longer than 2048 bytes"* — was a *slicing* panic in that exact
file, reachable from user source plus a user-supplied macro set. It was adopted
on replayable evidence rather than the weaker prevention footing the issue
assumed.

## Scope was decided by measurement

`arithmetic_side_effects` on the issue's suggested `src/metrics/` gives **313**
hits. The distribution is what decides it:

| Operator | Count |
|---|---|
| `+=` (counter increments) | 267 |
| binary `+` | 29 |
| **binary `-`** | **15** |
| binary `*` | 2 |

All 15 subtractions are in `src/metrics/loc/`, and `loc` is the only module in
`src/metrics/` that computes on tree-sitter *span coordinates* rather than its
own counters. The other 244 hits are not this bug class and would bury it.

**Total `#[allow]` count across both lints: 10**, each naming its invariant —
nine per-*function* in `c_macro.rs` (so a newly added function is covered by
default, verified with a throwaway unguarded index) and one file-level in
`line_set.rs`. Zero in the `loc` span code: 36 `logical_lines += 1` collapsed
into `Lloc::count_logical_line`, the rest made explicitly saturating.

The `line_set.rs` carve-out is deliberate and is the one worth a reviewer's
attention: its arithmetic is on bitset *word indices*, where saturating
`self.words[word - self.first_word]` to index 0 would read the wrong row and
silently miscount, whereas the current form panics on a corrupt offset — the
way `intersection_len`'s existing comment says it is meant to fail. Prevention
and masking point in opposite directions there.

## `unwrap_used` is not a `[workspace.lints]` entry

**0 production `unwrap()`, 37 production `expect()`** — measured over lib and
bin targets so `#[cfg(test)]` modules are excluded by construction. Across
`--all-targets` the same measurement is **1,023 `unwrap()` and 2,157
`expect()`**, essentially all legitimate test code, so a Cargo lint (which
applies to every target of its package) was not an option.

It is `#![cfg_attr(not(test), warn(clippy::unwrap_used))]` at each of the ten
production crate roots. The load-bearing fact — checked rather than assumed,
because the widely-repeated claim is the opposite — is that **`cfg(test)` *is*
set for integration-test crates**, not just the unit-test target. One form
therefore covers `#[cfg(test)]` modules and `tests/*/main.rs` alike: no
per-file carve-out, no footgun for the next integration test added, and **zero
`#[allow]`s**.

`expect_used` is declined. All 37 sites already name their invariant in the
message, the form `AGENTS.md` sanctions, so gating it buys 37 annotations that
restate the line above them. The distinction that decides it is not
`expect`-vs-`unwrap` but **what can invalidate the invariant**: the
`FEATURES_PINNED` sites rested on a `#[non_exhaustive]` enum a *dependency
bump* could change underneath them; these 37 rest on facts local to this
repository. That reasoning lives in `[workspace.lints.clippy]`, where the next
person will look for the absent lint.

## `FEATURES_PINNED` — wider than the issue scoped

#1152 named `dispatch.rs`. `big-code-analysis-web/.../handlers.rs` carried the
identical latent panic at 7 more sites, so all 15 are gone. Both crates already
had an error channel, which is the tell the `expect` was never necessary: the
CLI maps to `io::Error(InvalidData)` that the concurrent runner reports per file
and continues past; the web folds into the existing sanitized `500`, which is
the same status unwinding through `spawn_blocking` already produced — what
changes is that the log names the real cause and no worker thread unwinds.

Lessons-learned #26, which *recommended* the `.expect(FEATURES_PINNED)` pattern,
is corrected in the same change.

## Behaviour and risk

- **Metric values are unchanged.** A saturating operation equals the plain one
  unless it would have overflowed, and none does. No snapshots move.
- The `enums` codegen changes are verified by the existing
  `utils/check-enums-codegen-drift.sh`, which regenerates in both `rust` and
  `c_macros` modes and diffs against the checked-in output — `c_macros` is the
  mode using the `CARGO_MANIFEST_DIR` path that changed. Byte-identical.
- `dispatch_find` and `dispatch_exemptions` each gain one `nexits` — the `?`
  that replaced the `expect` — tripping the **soft** tier at 4.75. Both are
  baselined in the same commit per the refresh discipline; the hard limit of 5
  is unchanged and still met. This is the `?`-is-an-exit artifact `AGENTS.md`
  calls out, not new branching.

## Tests and coverage

Both new error paths are unreachable end-to-end by the feature pin, so they are
pinned by unit tests on the *mapping* rather than by a vacuous end-to-end test.
Every mutation against them is caught — wrong `ErrorKind`, cause stringified
instead of carried, log line deleted, wrong `ParseError` variant — each
confirmed to fail on an assertion rather than a compile error.

Patch coverage cannot drop: outside `enums/**` and `xtask/**` (both in
`codecov.yml`'s `ignore` list as code-generation helpers), the lint-gate diff
is 30 lines that are entirely comments and the `#![cfg_attr(...)]` attribute —
zero coverable lines, measured.

## Reviewer notes

- `make pre-commit` is green.
- Four rounds of quality skills ran over the branch (`simplify-rust`,
  `rust-optimize`, `review`, `audit-tests`) plus a final `/code-review`. Their
  findings are folded in; several were corrections to **my own** over-claiming
  comments, which is why four commits here are prose-only.
- Commit messages in this branch say "sixteen call sites"; the correct figure
  is fifteen plus two constants. `CHANGELOG.md` and lessons-learned are
  corrected; the commit messages are left as history.
- **#1228** is filed as a follow-up: `enums` is CI-linted with `-D warnings`
  but *without* the workspace pedantic set, since #158 scoped propagation to
  "the three shipping crates". 23 pedantic warnings, all pre-existing,
  deliberately not bundled here.

### DIFF
--- a/.bca-baseline.toml
+++ b/.bca-baseline.toml
@@ -169,9 +169,21 @@ value = 5.0
 
 [[entry]]
 path = "big-code-analysis-cli/src/dispatch.rs"
+qualified = "dispatch_exemptions"
+metric = "nexits"
+value = 5.0
+
+[[entry]]
+path = "big-code-analysis-cli/src/dispatch.rs"
 qualified = "dispatch_find"
 metric = "nargs"
 value = 6.0
+
+[[entry]]
+path = "big-code-analysis-cli/src/dispatch.rs"
+qualified = "dispatch_find"
+metric = "nexits"
+value = 5.0
 
 [[entry]]
 path = "big-code-analysis-cli/src/dispatch.rs"
@@ -1005,7 +1017,7 @@ value = 5.0
 path = "src/metrics/loc/perl.rs"
 qualified = "PerlCode::compute"
 metric = "halstead.effort"
-value = 55025.91689557041
+value = 52134.594881912846
 
 [[entry]]
 path = "src/metrics/loc/shared.rs"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ for historical reference.
 ### Fixed
 
 - The CLI and web crates no longer terminate on a library parse error.
-  All sixteen `.expect(FEATURES_PINNED)` call sites — eight in `bca`'s
+  All fifteen `.expect(FEATURES_PINNED)` call sites — eight in `bca`'s
   dispatch helpers, seven in `bca-web`'s handlers, plus the constant
   itself in each crate — now propagate onto the error channel each
   caller already had: an `io::Error` of kind `InvalidData` that the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,13 @@ for historical reference.
   costs nothing today and fails CI on the first production `unwrap()`
   added. `clippy::expect_used` is deliberately **not** enabled — all 37
   production `expect` sites already name their invariant in the message,
-  the form `AGENTS.md` sanctions (#1227).
+  the form `AGENTS.md` sanctions. The workspace-excluded `enums` codegen
+  crate carries the same gate: it is CI-linted by `make enums-check` but
+  invisible to `cargo clippy --workspace`, so its 7 production
+  `unwrap()` calls were outside the original count. One was a live
+  latent panic — `names.iter().map(..).max().unwrap()` in the Go
+  generator panics on a token-less grammar — and the rest now propagate
+  onto the `io::Result` each generator already returned (#1227).
 
 ## [2.1.0] - 2026-08-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,44 @@ for historical reference.
 
 ## [Unreleased]
 
+### Fixed
+
+- The CLI and web crates no longer terminate on a library parse error.
+  All sixteen `.expect(FEATURES_PINNED)` call sites — eight in `bca`'s
+  dispatch helpers, seven in `bca-web`'s handlers, plus the constant
+  itself in each crate — now propagate onto the error channel each
+  caller already had: an `io::Error` of kind `InvalidData` that the
+  concurrent runner reports per file and continues past, and the
+  existing sanitized `500` that logs the cause server-side. The pinned
+  `all-languages` feature does make `MetricsError::LanguageDisabled`
+  unreachable, but `MetricsError` is `#[non_exhaustive]` and documents
+  that variants may be added in a *minor* release, so the `expect` was a
+  panic scheduled against a routine dependency bump rather than an
+  invariant. No behaviour changes today: the only reachable outcome is
+  still success (#1152).
+
+### Changed
+
+- `clippy::arithmetic_side_effects` is enforced on the `loc` metric
+  module, and the span arithmetic there is now explicitly saturating.
+  `Loc` is the one metric computing on tree-sitter row coordinates, and
+  #1051 was a `usize` underflow of exactly that shape — a Rust doc
+  comment at EOF drove `end - 1` below zero from an input as small as
+  `/// x`, panicking in debug and wrapping to `usize::MAX` in release.
+  Validated by replaying the lint against the pre-#1051 tree, where it
+  flags both reported panic sites. Metric values are unchanged: a
+  saturating operation is identical to the plain one unless it would
+  have overflowed, and none does (#1152).
+- `clippy::indexing_slicing` is enforced on `src/c_macro.rs`, the C/C++
+  macro-masking byte lexer, with nine per-function carve-outs each
+  naming the bound that makes its indexing safe. Validated by replaying
+  it against the pre-#126 tree, where it flags the `&DOLLARS[..]` slices
+  that panicked on macro identifiers longer than 2048 bytes. The one
+  slice whose bound is established in a *different* function —
+  `step_raw_string`'s delimiter comparison, carried through
+  `LexState::RawString` — is hardened with `get` rather than allowed
+  (#1152).
+
 ## [2.1.0] - 2026-08-06
 
 A feature and correctness release on the `2.x` line, and the first to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,10 +75,12 @@ for historical reference.
   the form `AGENTS.md` sanctions. The workspace-excluded `enums` codegen
   crate carries the same gate: it is CI-linted by `make enums-check` but
   invisible to `cargo clippy --workspace`, so its 7 production
-  `unwrap()` calls were outside the original count. One was a live
-  latent panic — `names.iter().map(..).max().unwrap()` in the Go
-  generator panics on a token-less grammar — and the rest now propagate
-  onto the `io::Result` each generator already returned (#1227).
+  `unwrap()` calls were outside the original count. They now propagate
+  onto the `io::Result` each generator already returned, except the Go
+  generator's `max()` width, which becomes `unwrap_or(0)`. None was a
+  reachable crash: every one rests on an invariant as solid as the 37
+  `expect` sites left alone. The difference is that an `unwrap()` states
+  no invariant, which is the whole basis for gating it (#1227).
 
 ## [2.1.0] - 2026-08-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,18 @@ for historical reference.
   `step_raw_string`'s delimiter comparison, carried through
   `LexState::RawString` — is hardened with `get` rather than allowed
   (#1152).
+- `clippy::unwrap_used` is enforced on production code across every
+  crate, as `#![cfg_attr(not(test), warn(...))]` at each of the eight
+  lib/bin roots rather than a `[workspace.lints]` entry: a Cargo lint
+  applies to every target of its package, and the ban is a production
+  rule — this workspace has **0** production `unwrap()` calls against
+  1,023 legitimate ones in test targets. `cfg(test)` is set for
+  integration-test crates as well as the unit-test target, so the gate
+  needs no per-file carve-out and carries zero `#[allow]`s. Adopting it
+  costs nothing today and fails CI on the first production `unwrap()`
+  added. `clippy::expect_used` is deliberately **not** enabled — all 37
+  production `expect` sites already name their invariant in the message,
+  the form `AGENTS.md` sanctions (#1227).
 
 ## [2.1.0] - 2026-08-06
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -118,6 +118,25 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(chain_audit)'] }
 pedantic = { level = "warn", priority = -1 }
 module_name_repetitions = "allow"
 
+# `clippy::unwrap_used` is deliberately NOT declared here. It is set as
+# `#![cfg_attr(not(test), warn(clippy::unwrap_used))]` at each production
+# crate root instead (the eight lib/bin roots), because a Cargo lint
+# applies to every target of its package and the `unwrap` ban is a
+# production rule: this workspace's test targets hold 1_023 legitimate
+# `unwrap()` calls, against 0 in production. `cfg(test)` is set for
+# integration-test crates as well as for the unit-test target, so the
+# per-root form needs no per-file carve-out and carries zero `#[allow]`s.
+#
+# `clippy::expect_used` is not enabled at all. All 37 production `expect`
+# sites already name their invariant in the message, which is the form
+# `AGENTS.md` sanctions, so gating it would buy 37 annotations that each
+# restate the line above them. The distinction that matters is not
+# `expect`-vs-`unwrap` but *what can invalidate the invariant*: the
+# `FEATURES_PINNED` sites removed in #1152 rested on a `#[non_exhaustive]`
+# enum a dependency bump could change underneath them, whereas these 37
+# rest on facts local to this repository, which review and tests cover.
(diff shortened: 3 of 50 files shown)


dekobon:
## Patch coverage: the three missing lines are covered

Codecov reported **95.52% patch vs 98.33% project**, with three missing lines —
the `PreprocArg` range-insert in `loc/c.rs`, `loc/mozcpp.rs` and `loc/objc.rs`
(each file 66.66%: two `count_logical_line` lines covered, the range line not).

C++'s **identical** copy of that arm was already covered. That is exactly the
per-language blind spot `AGENTS.md` warns about — the four C-family `Loc`
modules are deliberate clones, so one passing language said nothing about the
other three.

`debd1538` adds `a_continued_macro_body_counts_every_row_it_spans`. A
backslash-continued `#define` body parses as a single `PreprocArg` node
spanning every continuation row, and each row is PLOC because
`tree-sitter-cpp` does not expand macros. The fixture sweeps all four
languages through `metrics_verbatim`, which takes a `LANG` — the only way to
reach `Mozcpp`, which owns no file extension and so gets no
integration-snapshot coverage at all.

Values are measured rather than guessed, and the test is confirmed
**discriminating per language**: deleting the arm from each module in turn
fails it, each perturbation confirmed to compile first so the failure is the
assertion and not the build.

| Module | arm deleted | result |
|---|---|---|
| `loc/c.rs` | `ploc` 4 → 3 | caught |
| `loc/mozcpp.rs` | `ploc` 4 → 3 | caught |
| `loc/objc.rs` | `ploc` 4 → 3 | caught |
| `loc/cpp.rs` | `ploc` 4 → 3 | caught |

That is stronger than the coverage number asks for: a line can be *executed*
without being *discriminated*, and these are both.

## Correcting my own earlier claim

I wrote in the PR body that the change has "zero coverable lines". That was
measured over the lint-gate commits only (`a0925a7f..HEAD`), not the whole PR
— the #1152 half adds real executable code. The accurate statement is the one
above: three coverable lines were uncovered, and they now are not.

Issue #1235, opened 2026-08-08
ci(fuzz): decide gating and report every failing target


dekobon:
Two follow-ups on `.github/workflows/fuzz.yml`, both known limitations
of what landed in #1154 rather than surprises. Filed together because
they are the same file and the same question: does a fuzz failure stop
anything, and does it tell you everything it found?

## 1. The workflow is advisory, not gating

`ci.yml` aggregates its jobs into a single `ci` check, and that is what
branch protection requires. `fuzz.yml` is a *separate workflow*, so it
is not in that aggregator's `needs:` and a red fuzz run does not block a
merge.

The separation is deliberate and should stay: cargo-fuzz needs nightly,
`ci.yml` sets `RUSTFLAGS: "-D warnings"` workflow-wide, and pairing the
two turns every new nightly lint into a red X on unrelated PRs — the
same reasoning the `coverage` job records for staying on stable.

So the fix is not to move the job. It is to add the `Build and smoke
targets` check to branch protection, which is a repository-settings
change and cannot be made from a PR.

Worth deciding rather than defaulting to: the per-PR budget is 10 000
runs per target over a committed corpus, which is a compile-and-smoke
gate, not a hunt. Making it *required* also makes every flake and every
nightly-toolchain hiccup a merge blocker. Requiring it is defensible;
so is leaving it advisory and watching the quarterly cron. What is not
defensible is the current state, where nobody has chosen.

## 2. `fuzz-smoke` stops at the first crashing target

The recipe loops over `cargo +nightly fuzz list` under `set -e`, so the
first target that fails aborts the loop and the remaining targets never
run:

```make
for t in $$(cargo +nightly fuzz list); do \
  ... cargo +nightly fuzz run "$$t" ... ; \
done
```

Fine for a PR gate — one crash is enough to stop the merge — but wrong
for the quarterly cron, which is the run nobody is watching and the one
that most wants a complete picture. As it stands a cron run that finds
three unrelated crashes reports one, and the next run rediscovers the
other two only after the first is fixed.

Fix: collect per-target status, keep going, and fail at the end with the
list. Keep the artifact upload as-is — it already globs `fuzz/artifacts/`
across targets, so the reproducers for later failures would be captured
the moment the loop stops discarding them.

## Done when

- [ ] A decision is recorded on whether `Build and smoke targets` is a
      required check, and branch protection matches it.
- [ ] `make fuzz-smoke` runs every target regardless of earlier
[...]

Issue #1234, opened 2026-08-08
chore(release): gate excluded-crate lockfile staleness


dekobon:
Found while landing the fuzz crate (#1154), but the `enums` half is
pre-existing and would have fired at the next release.

## The defect

A workspace-excluded crate carries its own `Cargo.lock`, and
`cargo update --workspace` — the command `RELEASING.md` calls
**mandatory** during a version bump — does not reach it. Both excluded
lockfiles record versions that a bump invalidates:

- `enums/Cargo.lock` records `enums` at its own `[package] version` and
  the five vendored grammars at their `=X.Y.Z` pins.
- `fuzz/Cargo.lock` records `big-code-analysis` at the version it
  resolved against.

Every gate that consumes them passes `--locked`, so the next bump breaks
them:

```text
error: cannot update the lock file .../enums/Cargo.lock
       because --locked was passed to prevent this
```

Both verified by editing the version and re-running, not by inspection:

| Edit | Command | Result |
|---|---|---|
| root `[workspace.package] version` 2.1.1 → 2.2.0 | `cargo +nightly test --manifest-path fuzz/Cargo.toml --locked` | fails as above |
| `enums/Cargo.toml` `[package] version` 2.1.1 → 2.2.0 | `cargo test --manifest-path enums/Cargo.toml --locked` | fails as above |

`make enums-check` is wired into `make lint`, `pre-commit` and `ci`, so
the `enums` case turns the whole local gate red on a release-prep commit
that is otherwise correct — and it looks like a defect in whatever else
that commit touched.

## What is already done

`RELEASING.md` now names the step, next to the existing
`cargo update --workspace` paragraph:

```bash
cargo update --manifest-path fuzz/Cargo.toml --workspace
cargo update --manifest-path enums/Cargo.toml --workspace
```

## Why that is not enough

It is a checklist item guarding a failure whose whole character is
"someone forgot a checklist item", and it fails *late* — not at
`cargo publish`, since neither crate is published, but as a red gate on
a tree that reads as correct.

The lockstep version policy already has a gate
(`utils/check-versions.py`), and that gate does not know these lockfiles
exist: `EXCLUDED_LEAF_DIRS` and `INTERNAL_PIN_MANIFESTS` enumerate
manifests only.

## Proposal

Extend `utils/check-versions.py` to assert each excluded crate's
`Cargo.lock` agrees with its manifest — that the `[[package]]` entry for
the crate itself, and for any path dependency on a workspace crate,
[...]

Issue #1233, opened 2026-08-08
fix(deps): tree-sitter-perl leaks its heredoc queues


dekobon:
Found by the `parse_perl` target on its first bounded run, while landing
the fuzz crate (#1154).

## The defect

`tree_sitter_perl_external_scanner_destroy` frees the `Scanner` and
nothing else. Every other free in it is present but commented out
(`scanner.c:1225`, tree-sitter-perl 1.1.2):

```c
void tree_sitter_perl_external_scanner_destroy(void *payload) {
  Scanner *scanner = (Scanner *)payload;
  // deleteBoolQueue(scanner->heredoc.heredoc_allows_indent);
  // deleteBoolQueue(scanner->heredoc.heredoc_allows_interpolation);
  // deleteStringQueue(scanner->heredoc.heredoc_identifier_queue);
  // for (size_t i = 0; i < scanner->heredocs.len; i++) {
  //     Heredoc *heredoc = &scanner->heredocs.data[i];
  //     STRING_FREE(heredoc->current_leading_word);
  //     STRING_FREE(heredoc->delimiter);
  // }
  // VEC_FREE(scanner->heredocs);
  free(scanner);
}
```

So the three `create_queue()` allocations `heredoc_new` makes
(`scanner.c:146`, called from `:364-366` via
`tree_sitter_perl_external_scanner_create` at `:1198`) leak, along with
whatever the queues and the heredoc vector still hold at destroy time.

LeakSanitizer stack, reached from ordinary safe Rust:

```text
malloc
  create_queue                                    scanner.c:146
  heredoc_new                                     scanner.c:364
  tree_sitter_perl_external_scanner_create        scanner.c:1198
  ts_parser__external_scanner_create              parser.c:369
  ts_parser_parse                                 parser.c:2103
  big_code_analysis::node::parser_cache::parse_on_scratch_parser
  big_code_analysis::spaces::Ast::parse
```

## It is bounded, and that matters

The code reads like a per-file leak. It is not, and nothing should be
escalated on the assumption that it is:

| Run length | Leaked |
|---|---|
| `-runs=3000` | 3 144 bytes in 9 allocations |
| `-runs=30000` | 3 144 bytes in 9 allocations |

Identical, because `node::parser_cache::parse_on_scratch_parser` reuses
one thread-local `Parser`, so tree-sitter creates the external scanner
once per thread per language rather than once per file. The cost is a
few kilobytes per worker thread for the process lifetime, not growth
under load.

[...]

PR #1251, opened 2026-08-08
test(fuzz): add a targeted cargo-fuzz crate

### LINKED ISSUE #1154: test(fuzz): add a targeted cargo-fuzz crate
Split from #1053 (§2). See that issue for the threat model and the explicit
Scorecard non-goal — in particular, **do not** satisfy the Scorecard
`FuzzingID` check cosmetically. Its Rust detector is the literal substring
`libfuzzer_sys` in any `*.rs` file, so a throwaway file turns it green with
zero security benefit. Land the substance; the badge follows.

## Summary

Add a targeted `cargo-fuzz` crate. Justified by coverage gaps, not by the
Scorecard alert.

## The justification has changed — re-read before starting

#1053 argued for this on the strength of two live availability bugs, #1051
(5-byte input panics the library in release) and #1052 (kilobyte-scale CPU
exhaustion, unauthenticated on `bca-web`). **Both are now fixed**, along with
#1054, #1056, #1059 and #1060. So the case is no longer "a fuzzer would have
caught these two"; it is purely prospective.

That is still a reasonable case, but it is a different one, and it should be
scoped around **what the static lints in the sibling split issue cannot
catch** rather than inherited wholesale. `clippy::arithmetic_side_effects`
would have found #1051 statically; a fuzzer is for the shapes no lint can see.

## Counter-evidence worth keeping

A 112-file adversarial byte corpus — invalid UTF-8, NUL bytes, UTF-8/UTF-16
BOMs, mixed EOL, unterminated constructs, 2 MB single tokens — across 7
subcommands found **zero** panics in the encoding/IO layer. That path is
genuinely hardened, and blanket byte fuzzing would mostly re-confirm work
already done. Target the gaps instead. Do not re-litigate this.

## Scope

- [x] Add a `fuzz/` cargo-fuzz crate, excluded from the workspace so
      `cargo test --workspace` stays unaffected.
- [x] Target `analyze(Source::from_bytes(lang, data))` **directly**, one target
      per language or an `Arbitrary`-selected language. This must bypass
      `normalize_line_endings` — normalizing first is exactly what hid #1051,
      and a harness that normalizes will pass vacuously.
- [x] Add targets for `ops`, `dump`, and `comment_rm`.
- [x] Add a depth-oriented structured generator for the #1052 complexity class;
      a naive byte mutator finds it far more slowly than a generator that knows
      how to nest.
- [x] CI: `cargo +nightly fuzz build` as a cheap compile gate on every PR, plus
      bounded smoke runs. Use `-runs=N`, **not** `-max_total_time` — wall-clock
      bounds make failures non-reproducible across runner speeds. Set
      `-timeout=10` so complexity regressions surface as failures.
- [x] Commit seed corpora. `cargo fuzz init` writes a `fuzz/.gitignore` that
      excludes `corpus/`, so seeds need `git add -f` or an amended ignore file.
- [x] Every crasher becomes both a committed corpus seed **and** a normal
      `#[test]`, so the regression is enforced without running the fuzzer.

Nightly is already a project dependency (`cargo +nightly udeps` is in the
validation gate), so the usual "fuzzing needs nightly" objection is paid for.

## Define "done" before starting

Coverage-guided fuzzing has no natural stopping point, and without a stated
one this issue stays permanently open. Proposed acceptance: `fuzz/` exists,
builds in CI, ships seed corpora, and has either found ≥1 real bug or returned
clean over an agreed run budget — at which point the issue closes and further
fuzzing is maintenance, not project work.

## Sequencing

Do this **last** of the three #1053 splits: after the lints (cheapest, catches
the known class statically) and after the CLI resource bounds (largest
remaining exposure).

## Assessment

| Dimension  | Rating |
|------------|--------|
| Difficulty | Medium |
| Complexity | Medium |
| Priority   | Low |

**Difficulty — Medium.** A new crate, per-language targets, a structured
generator for the depth class, CI wiring, and seed corpora — none individually
hard, but a lot of surface for a first landing.

**Complexity — Medium.** A workspace-excluded crate plus CI workflows. Nothing
downstream depends on it and no product behaviour changes, which keeps it from
being higher.

**Priority — Low.** Every bug it was filed to systematize is now fixed, the
encoding/IO layer was separately shown clean against a 112-file adversarial
corpus, and the known-reachable class is better caught statically by #1152.
Genuinely valuable as prevention, but last of the three #1053 splits.

---

## Resolution Plan

Not started; no `fuzz/` directory on `fix/batch-2026-07-31-2`. Sequenced
**last** of the three #1053 splits, behind #1152 (lints) and #1153 (CLI
bounds).

### Re-argue the case before writing code

The Summary already records that the motivating bugs are fixed and the case is
now prospective. Make that concrete before starting, because it determines the
target list:

- **#1152 is the control.** Whatever `arithmetic_side_effects` and
  `indexing_slicing` catch statically does not need a fuzz target. Run #1152's
  validation pass first and let its `#[allow]` list define the residue — every
  allowed site is arithmetic a human asserted is bounded, which is exactly the
  population worth fuzzing.
- **The 112-file adversarial corpus found zero panics in the encoding/IO
  layer.** That path is hardened; do not re-fuzz it. The targets below are the
  parse-and-walk layer, which that corpus exercised only incidentally.

### Steps

1. `cargo fuzz init` into a workspace-**excluded** `fuzz/` crate. Amend the
   generated `fuzz/.gitignore` so `corpus/` is tracked, or the seeds silently
   never land.
2. Target `analyze(Source::from_bytes(lang, data))` **directly**. This is the
   single most important design point: it must bypass `normalize_line_endings`,
   because normalizing first is what hid #1051, and a harness that normalizes
   passes vacuously while looking like coverage.
3. One target per language, or an `Arbitrary`-selected language. Prefer
   per-language targets — a shared one spends most of its budget on whichever
   grammar is cheapest to reach.
4. Targets for `ops`, `dump`, and `comment_rm`.
5. A depth-oriented structured generator for the #1052 complexity class. A
   naive byte mutator reaches deep nesting far more slowly than a generator
   that knows how to nest, and #1052's class is the one a fuzzer is uniquely
   suited to.
6. CI: `cargo +nightly fuzz build` as a compile gate on every PR, plus bounded
   smoke runs with `-runs=N` — **not** `-max_total_time`, which makes failures
   non-reproducible across runner speeds — and `-timeout=10` so a complexity
   regression surfaces as a failure rather than a slow run.
7. Every crasher becomes a committed corpus seed **and** a normal `#[test]`.
   The test is the durable half: it enforces the regression on every
   `cargo test`, without anyone running the fuzzer again.
8. Record the stopping rule in the PR and close on it. Coverage-guided fuzzing
   has no natural end, and an issue with no definition of done stays open
   forever.

### On the Scorecard non-goal

Restating because this is the change that will turn the badge green and the
temptation arrives with it: the check's Rust detector is the literal substring
`libfuzzer_sys` in any `*.rs` file, with no Cargo parsing and no requirement
that anything runs. Do not let that shape any decision here — in particular, do
not keep a target that finds nothing merely because deleting it would drop the
score.

## Assessment

Ratings unchanged from the table above (Medium / Medium / Low).

---

## Status: implemented on `fix/1154-cargo-fuzz`, sweep pending

Four commits. `make pre-commit` green.

**Delivered.** Ten libFuzzer targets in a workspace-excluded `fuzz/`
crate (eight per-language, plus `preproc_macro` and `nested_depth`), 89
committed seeds, `make fuzz-check` / `fuzz-smoke` / `fuzz-run`, a
separate `.github/workflows/fuzz.yml`, and
`docs/development/fuzzing.md`.

**Deviations from the scope above, each with a reason.**

- *Target breadth.* Eight per-language targets, not one per `LANG`
  variant. Chosen on stated evidence rather than by enumerating the
  enum; the languages left out are listed with reasons in the doc so the
  omission reads as a decision.
- *`ops` / `dump` / `comment_rm`.* Not separate targets — they are seams
  on an already-parsed `Ast`, so every target runs them through one
  shared fan-out instead of re-paying the parse three times.
- *CI placement.* A separate workflow, not a `ci.yml` job: that workflow
  sets `RUSTFLAGS: "-D warnings"` globally and pairing it with nightly
  red-Xes CI on unrelated new nightly lints. It is therefore **not** in
  the `ci` aggregator's `needs:` and is advisory until branch protection
  names it.
- *"Every crasher becomes a seed and a `#[test]`."* Honoured, with one
  exception written into the doc: the finding was a *leak*, which
  `cargo test` cannot observe outside a sanitizer build. Its durable
  artifact is a documented entry in `fuzz/lsan-suppressions.txt`.

**Found.** A real but *bounded* leak in tree-sitter-perl 1.1.2 (queue
frees commented out upstream), plus two defects in our own code — a
`LANG::C` preprocessor test that passed with the arm it covers disabled,
and a first seed corpus whose twelve entries all decoded to one language.
See the comment below for the measurements.

**Remaining before this closes.** The agreed stopping rule is
`-runs=1000000` per target; that sweep is running against the final
binaries. The issue stays open until it reports clean.

dekobon:
Adds a targeted `cargo-fuzz` crate. Closes #1154.

## Scope, and why it is narrow

The bugs that motivated #1053 are all fixed, so the case is prospective
and the issue asks for it to be scoped around **what the static lints in
#1152 cannot catch** rather than inherited wholesale. That residue is
small and concrete: nine per-function `#[allow(clippy::indexing_slicing)]`
sites in `src/c_macro.rs`, each a computed index into attacker-controlled
bytes whose bound is asserted by a human comment. Separately, a 112-file
adversarial corpus already showed the encoding/IO layer clean, so this
does not re-fuzz it.

Ten targets: eight per-language, chosen on stated evidence rather than by
enumerating `LANG`; `preproc_macro` for the residue above; and
`nested_depth`, a structured generator for the deep-nesting class.
Languages deliberately left out are listed with reasons in the doc, so
the omission reads as a decision.

## Design points that are load-bearing

- **Bytes reach the parser unnormalised.** Targets use
  `Source::from_bytes`. The file-reading path appends a trailing newline,
  which is what made a node ending at EOF unreachable and hid #1051 — a
  harness that inherited it would pass vacuously while looking like
  coverage. `Ast::from_path` normalises and no target uses it.
- **One parse, every walk.** `ops`, `dump` and comment removal are seams
  on an already-parsed `Ast`, not separate entry points, so each target
  parses once and fans out. It also serialises the results, which is the
  only thing that reaches the recursion #1056 bounded.
- **The C grammars are ASan-instrumented too.** It has to be the
  triple-qualified `CFLAGS_<triple>` spelling: `cc` picks the `HOST_`
  prefix over `TARGET_` when not cross-compiling, so `TARGET_CFLAGS` is
  silently inert for a native fuzz build. Verified with `nm` on the built
  scanner objects. This is the configuration that found the
  tree-sitter-cpp `deserialize` overflow (#1058).

## Non-vacuity was measured

`c_macro::replace` runs only when `Source::preproc` is set, so no
per-language target reaches it. Verified by perturbation: a `panic!` at
the top of `replace` fires for `preproc_macro` and does **not** fire for
`parse_cpp` on the same input.

The first version of that probe was itself wrong — it matched the panic
string in the combined output of `cargo fuzz run`, and rustc's
`unreachable_code` diagnostic quotes the panic line, so both targets
"reached" the function and the probe reported a conclusion it had not
measured. Building first and running the binary directly fixed it.

## What it found

**A real leak in tree-sitter-perl 1.1.2** (#1233):
`external_scanner_destroy` frees the `Scanner` and nothing else — the
three `create_queue()` frees, the per-heredoc `STRING_FREE`s and the
`VEC_FREE` are all present but commented out upstream.

It is **bounded, not per-parse**, and the distinction is in the issue
because the code reads like the worse thing: 3 144 bytes in 9
allocations, *identical* at `-runs=3000` and `-runs=30000`, because the
parser cache creates one scanner per thread per language. Not the #1052
availability class. Suppressed by symbol so other leaks on that path
still fail.

## Defects fixed in our own code along the way

- **The `LANG::C` preprocessor arm had no test that could observe it.**
  #721 added `parse_then_metrics_c_with_preproc_…` to give that arm
  "direct coverage" and it could not: its only assertion is a
  `cyclomatic_sum`, and `DBG ? x : 0` and `$$$ ? x : 0` have the same
  complexity, so it passed unchanged with the whole arm disabled. Fixed,
  plus a new end-to-end test — nothing in the workspace chained the
  public `preprocess` into `Ast::parse`.
- **`AstCfg::comment` is a suppression flag** and the harness passed
  `true`, dumping every tree with comment nodes dropped — the opposite of
  intent, given each seed carries a comment on purpose.
- **The first `nested_depth` seed corpus was vacuous.** All twelve seeds
  decoded to `Rust`; a 16-million-combination search could not produce a
  single non-Rust input, because the `Arbitrary` derive's byte layout is
  the crate's business. Replaced with a hand-written layout, so seeds are
  writable and `seeds_cover_every_language` can assert them.
- **`fuzz/` is invisible to `cargo fmt --all`**, so the crate landed
  unformatted with no gate able to say so. `fuzz-check` now has a fmt
  stage.
- **git was about to rewrite the byte-exact seeds.** `git add` warned it
  would normalise CRLF to LF, which would have silently replaced those
  tests with different ones. The corpus is `-text -diff`.
- **libFuzzer writes into the first corpus directory it is given.**
  Pointed at the committed seeds — cargo-fuzz's default — one 5 000-run
  pass grew 93 checked-in files to over 4 000. The recipes now pass a
  gitignored work directory first.

## Deliberately not done

- **Not wired into `make pre-commit`.** A nightly ASan build of every
  grammar is not a per-commit cost — the same call `chain-audit` and the
  benchmark harness make.
- **A separate workflow, not a `ci.yml` job.** `ci.yml` sets
  `RUSTFLAGS: "-D warnings"` workflow-wide and has no nightly job
  precisely because pairing the two red-Xes CI on unrelated new nightly
  lints. The consequence is stated plainly in the file and the doc: this
  workflow is **not** in the `ci` aggregator's `needs:`, so it is
  advisory until branch protection names it. Tracked in #1235 as a
  decision to make rather than a default to inherit.

## Follow-ups filed

- #1233 — the tree-sitter-perl leak (upstream; the suppression's removal
  is its close condition).
- #1234 — excluded-crate lockfile staleness has no gate. `fuzz/` and
  `enums/` both break `--locked` after a version bump; the `enums` half
  is **pre-existing** and would have turned `make lint` red on the next
  release-prep commit. `RELEASING.md` now names the step, but a checklist
  item is thin cover for a forgot-the-checklist failure.
- #1235 — fuzz workflow gating, and `fuzz-smoke` stopping at the first
  crashing target.

## Validation

`make pre-commit` green (`BCA_GATE: pass`). `make fuzz-check` green:
fmt, clippy, 4 crate tests, and all ten targets build. Coverage is
structurally unaffected — a workspace-excluded crate is unreachable from
`cargo llvm-cov --workspace`, and `fuzz/**` is in `codecov.yml`'s ignore
list as well.

**The `-runs=1000000` stopping rule is not yet met.** The issue requires
it before closing, and the sweep restarts against these binaries. Worth
knowing for anyone re-running it: the per-language targets manage
300–750 exec/s under ASan, but `nested_depth` runs at ~50 and slows as
libFuzzer finds deeper inputs, so that target alone is several hours. The
CI budgets are set at 10 000 runs per target per PR and 200 000 on the
quarterly cron, not at the local sweep's figure.


dekobon:
Filed #1288 for the `fix_includes` gap the review deferred, so it is tracked rather than lost with the branch.

Issue #1116, opened 2026-07-30
perf(cli): diff --since round-trips both sides through temp JSON files


dekobon:
## Summary

`bca diff --since` computes both sides' metrics in-process and then round-trips them through the filesystem: `walk_metric_set` (`big-code-analysis-cli/src/lib.rs:428-465`) runs a full walk with `output_dir: Some(json_out_dir)` — one JSON document per source file, through the unbuffered writer of #1115 plus a `create_dir_all` per file — then `load_dir_set` / `read_json` (`metric_diff.rs:534-565`) immediately re-walks the temp directory, re-opens, re-reads, and re-parses every document into `serde_json::Value` (`extract_metrics` additionally clones the metrics sub-object).

The `FuncSpace` trees are already in memory in the workers when this starts. Serializing to N temp files and parsing back into dynamically-typed `Value` doubles the IO and adds a full JSON parse + `Value` allocation per file, per side — the dominant cost of `bca diff --since` on any tree of size.

## Resolution Plan

1. Add an in-memory mode to `walk_metric_set`: reuse the existing aggregate-channel machinery (`lib.rs:146`, `walk.rs:15-21`) to stream per-file results back and build the `MetricSet` directly.
2. Keep the JSON-directory path for the explicit two-directory `bca diff <old> <new>` form (that form's input genuinely is a directory of documents).
3. Decide the comparison representation: if the differ stays `Value`-based, build `Value`s via `serde_json::to_value` in memory (no file IO, no parse); a typed comparison is a larger follow-up.
4. Existing `diff_since` integration tests (`big-code-analysis-cli/tests/diff_since.rs`, 22 tests) pin the user-visible output; they should pass unchanged.
5. Quote before/after wall time for `bca diff --since` on a real repo in the PR.

## Assessment

| Dimension  | Rating |
|------------|--------|
| Difficulty | Medium |
| Complexity | Medium |
| Priority   | Medium |

---

## Decision (2026-07-31)

**Keep the differ `Value`-based.** Settles step 3 of the Resolution Plan.

Build the `Value`s in memory via `serde_json::to_value` — no file IO, no parse,
no serialize/deserialize round-trip. A typed comparison is the larger, separate
follow-up the step already anticipates, and folding it in here would turn a
contained IO removal into a differ rewrite while 22 integration tests are
pinning the user-visible output.

Steps 1, 2, 4 and 5 are unchanged: in-memory mode for `--since`, keep the
[...]

PR #1377, opened 2026-08-30
perf(metrics): bound the cognitive nesting map

### LINKED ISSUE #1375: perf(cli): peak RSS scales with jobs × largest file (5.1 GB over a 100k-file tree)
## Summary

`bca metrics -O json` over a 100k-file tree peaks at **5.1 GB RSS** at
`--jobs 16`. The peak scales with the worker count and with the largest
files each worker is holding, not with the size of the tree: a single
~2.5 MB source file costs a worker 130–220 MB, roughly 50–85× the input.
Nothing documents this sizing rule, and `--jobs auto` picks the CPU
count with no regard to it, so a CI container that fits the *tree* can
still be OOM-killed by a handful of large files landing on workers at
the same time.

Observed while measuring #1158 (closed won't-fix on the same numbers —
the path list the runner materializes is ~30 MB at this scale, noise
against this peak).

## Evidence

Release `bca 2.2.1`, 16 cores, NVMe, `/usr/bin/time -f '%e %M'`. Every
run is `--no-config -O json > /dev/null`.

| run | wall | peak RSS |
|---|---|---|
| `tests/repositories/DeepSpeech` (12.7k files), `-j 1` | 19.9 s | 125 MB |
| same, `-j 2` | 12.3 s | 178 MB |
| same, `-j 16` | 3.4 s | 488 MB |
| `~/.cargo/registry/src` (100,476 files, 62,366 `.rs`), `-j 16` | 23.8 s | **5.1 GB** |

Per-file cost, `-j 1`, `--no-skip-generated` (all three are skipped as
generated by default; the registry run above was *default*, so its peak
came from files the generated-file heuristic did not catch):

| file | size | peak RSS |
|---|---|---|
| `crc-fast-1.10.0/src/tables.rs` | 2.77 MB | 137 MB |
| `encoding_rs-0.8.35/src/data.rs` | 2.58 MB | 217 MB |
| `brotli-8.0.4/src/enc/static_dict_lut.rs` | 1.87 MB | 134 MB |

Baseline process RSS with no analysis is ~12 MB, so those are the
working set of one file: tree-sitter tree, `FuncSpace` tree, Halstead
maps, and the rendered document, all alive together at the emit point.

The DeepSpeech ladder (125 → 178 → 488 MB for 1 → 2 → 16 workers) is
~25 MB per additional worker on ordinary-sized files; the registry's
5.1 GB / 16 workers is ~300 MB per worker, i.e. the large-file cost
above, retained.

## What has not been established

This is a measurement, not an attribution. Before anyone sizes a fix:

- Split the per-file working set between the tree-sitter tree, the
  `FuncSpace` tree, the metric maps, and the serialized document.
  Whichever dominates decides whether the fix is "drop the tree before
  serializing", "stream the document", or "nothing to do".
- Establish how much of the *process* peak is retention rather than
  live data. Each worker thread frees its trees per file, but glibc
  per-thread arenas keep freed pages, so peak RSS may be the *sum of
  each worker's worst file* rather than the worst simultaneous set.
  `MALLOC_ARENA_MAX=1` or a jemalloc run would show the difference.
- Whether the `OrderedStdout` reorder buffer (#1303) contributes: it
  holds finished documents while an earlier, larger file is still
  being analyzed, which is exactly the shape a large file creates.
  `--output-dir` vs stdout over the same tree separates it.

## Possible outcomes, none chosen

- Document the sizing rule (`jobs × ~80 × largest source file`) in the
  book's CI recipe so a container can be sized deliberately.
- Make `--jobs auto` consider available memory, or add a
  `--max-file-size` skip with a warning the way the generated-file
  skip already works.
- Reduce the per-file working set at the emit point, if the
  attribution shows one component dominates.

## Assessment

| Dimension  | Rating |
|------------|--------|
| Difficulty | Medium |
| Complexity | Medium |
| Priority   | Medium |

Medium rather than Low because the failure mode is an OOM kill in CI
with no diagnostic, and the trigger (a few large files in a large tree)
is ordinary.

dekobon:
Fixes #1375.

## What the peak actually was

#1375 was filed from a measurement with the attribution explicitly not
done, and two of its three hypotheses turned out to be wrong. The 5.1 GB
is not worker fan-out over ordinary files: the registry carries a dozen
`tree-sitter-*` crates whose generated `src/parser.c` runs 8–28 MB and
slips past the generated-file heuristic (the header opens with
`#include "tree_sitter/parser.h"`, no marker). At `-j 1` the whole
100k-file tree peaks at 1.31 GB — exactly the largest single file, so
nothing accumulates across files.

Splitting one 28 MB file's 1,265 MB by selection:

| selection (`-j 1`) | peak RSS |
|---|---|
| `dump` — parse only | 722 MB |
| `--metrics cyclomatic` / `halstead` / `abc` / `nargs` / `nexits` | 723–725 MB |
| `--metrics cognitive` | **1,264 MB** |
| all metrics, stdout | 1,265 MB |
| all metrics, `--output-dir` | 1,266 MB |

So ~720 MB is the parse and **cognitive alone added ~540 MB**. The
reorder buffer and the wire clone are ruled out by the last two rows
(`--output-dir` has neither and matches stdout); `bca check`, which has
no reorder buffer at all, peaked at 4.5 GB on the same tree.
`MALLOC_ARENA_MAX=1` gives 3.4 GB at 5× the wall time — real, but
secondary and not ours.

## The change

`NestingMap` was reserved at `descendant_count` and never freed a slot,
so it converged on one entry per node. Each slot is now freed by its
last reader, and the map grows from empty, so the live set is the
seeded-but-unvisited nodes — bounded by the traversal stack.
`propagate_nesting_to_children` reserves per sibling set so a wide
generated table still costs one allocation.

| run | before | after |
|---|---|---|
| objc `parser.c` (28 MB), `-j 1` | 1,265 MB / 5.1 s | **735 MB / 3.3 s** |
| cargo registry (100k files), `-j 16` | 5.1 GB | **3.6 GB** |
| DeepSpeech, `-j 1` / `-j 16` | 125 / 488 MB | 103 / 439 MB |

`make bench-scaling` unchanged, all 27 probes linear.

## Why this needs its own guard

No metric value moves — the retained entries were already dead — so
nothing in the output can distinguish the fix from its absence. Two
guards, deliberately overlapping:

- `walk_frees_every_cognitive_nesting_slot` reads the
  `nesting_slots_retained` counter over the four paths that seed a slot:
  a plain nested walk, the two overrides that write a child's slot ahead
  of the walk (Python comprehension clauses, Tcl `try` handler bodies),
  and an `exclude_tests` prune. Both freeing sites were verified by
  reverting them individually and watching only this test fail.
- A `debug_assert!` generalises that to every walk the suite runs —
  twenty languages of corpus fixtures rather than the four the test
  names. This matters because the freeing rule now has two owners, and a
  third `continue` added later would otherwise leak in silence.

## Incidental

- `Cognitive::SEEDS_NESTING` and `Node::descendant_count` existed only
  to size the reserve; removed with their tests.
- `parser_cache`'s module doc asserted a per-thread retention bound of
  "tens of KiB". The pinned tree-sitter caps those pools at fixed counts
  (`TS_MAX_TREE_POOL_SIZE`, `MAX_NODE_POOL_SIZE`) and the surrounding
  scratch arrays track parse depth rather than input size, so the figure
  was unsupported in both directions. The doc now states what the source
  actually shows.
- The CI recipe gains the `--jobs` memory sizing rule: `bca` is
  cgroup-CPU-aware but not memory-aware, and a parse-only run measures
  25–70× the source size.

## Deliberately not done

A `--max-file-size` gate, a memory-aware `--jobs auto`, a bounded
`OrderedStdout`, and an allocator swap all address terms the measurement
shows are small. The remaining ~720 MB on a 28 MB input is the parse
tree itself.

### DIFF
--- a/big-code-analysis-book/src/recipes/ci.md
+++ b/big-code-analysis-book/src/recipes/ci.md
@@ -869,11 +869,20 @@ Applies regardless of provider:
   macOS/Windows — so CI runners no longer need to thread
   `--jobs "$(nproc)"` through every recipe. `--jobs 1`
   remains a debugging knob, not a default. It is *not*
-  memory-aware: each worker holds one whole file in memory at a
-  time, and a parse-only run measures 25–70× the source size, so
-  peak RSS is roughly `jobs × 70 × the largest source file` in
-  the tree. A memory-limited container with many vCPUs should
-  pass an explicit `--jobs`.
+  memory-aware. Each worker holds one whole file at a time and a
+  parse-only run measures 25–70× the source size, so the
+  *in-flight* term is roughly `jobs × 70 × the largest source
+  file` in the tree — and `--jobs` bounds that term only.
+  Whether it is the whole story depends on the destination:
+  `--output-dir` writes each document as it is produced and
+  `bca check` reduces each file to its violations, so both stay
+  at the in-flight bound, while `metrics --output <FILE>`
+  collects every file's result and serializes it after the walk,
+  and structured stdout holds finished documents while an
+  earlier file is still being analyzed — which is worst when one
+  large file sorts early. A memory-limited container with many
+  vCPUs should pass an explicit `--jobs`, and prefer
+  `--output-dir` to `--output <FILE>` on a large tree.
 - **Always pass `--strip-prefix "$PWD/"` to `bca report markdown`**
   so the path column is identical across runners with different
   workspace paths. Without it the diff between two reports is
--- a/big-code-analysis-book/src/recipes/ci.md
+++ b/big-code-analysis-book/src/recipes/ci.md
@@ -868,7 +868,12 @@ Applies regardless of provider:
   cgroup-/cpuset-/quota-aware on Linux, OS CPU count on
   macOS/Windows — so CI runners no longer need to thread
   `--jobs "$(nproc)"` through every recipe. `--jobs 1`
-  remains a debugging knob, not a default.
+  remains a debugging knob, not a default. It is *not*
+  memory-aware: each worker holds one whole file in memory at a
+  time, and a parse-only run measures 25–70× the source size, so
+  peak RSS is roughly `jobs × 70 × the largest source file` in
+  the tree. A memory-limited container with many vCPUs should
+  pass an explicit `--jobs`.
 - **Always pass `--strip-prefix "$PWD/"` to `bca report markdown`**
   so the path column is identical across runners with different
   workspace paths. Without it the diff between two reports is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,24 @@ for historical reference.
 
 ## [Unreleased]
 
+### Performance
+
+- The metric walk's cognitive nesting map no longer grows to one entry
+  per AST node (#1375). Each node's slot was seeded by its parent, read
+  once by its own `compute`, read once more when its children were
+  seeded, and then kept until the walk ended; the map was also reserved
+  at the tree's node count up front. On a 28 MB generated
+  `tree-sitter` `parser.c` that reserve was 540 MB of a 1,265 MB peak.
+  The slot is now freed by its last reader, so the map holds only the
+  seeded-but-unvisited nodes — bounded by the traversal stack, not the
+  tree. Same file, release build, `-j 1`: 1,265 MB → 737 MB peak RSS
+  and 5.1 s → 3.9 s; `metrics -O json` over a 100k-file cargo registry
+  at `-j 16`: 5.1 GB → 3.6 GB; `tests/repositories/DeepSpeech` at
+  `-j 16`: 488 MB → 439 MB. No metric value moves. What remains is
+  dominated by the parse rather than by the walk: a parse-only
+  `bca dump` over the same files peaks at 25–70× the source size, which
+  the CI recipe now states as the `--jobs` memory sizing rule.
+
 ### Changed
 
 - The `tree-sitter` runtime is `=0.26.13`, up one upstream patch
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,9 @@ for historical reference.
   `-j 16`: 488 MB → 439 MB. No metric value moves. What remains is
   dominated by the parse rather than by the walk: a parse-only
   `bca dump` over the same files peaks at 25–70× the source size, which
-  the CI recipe now states as the `--jobs` memory sizing rule.
+  the CI recipe now records as the in-flight term in its `--jobs` sizing
+  guidance — alongside the destinations that add a result-set term on
+  top of it (`metrics --output <FILE>`, structured stdout).
 
 ### Changed
 
--- a/src/macros/mod.rs
+++ b/src/macros/mod.rs
@@ -42,10 +42,6 @@ macro_rules! implement_metric_trait {
     (Cognitive, $($code:ident),+) => (
         $(
            impl Cognitive for $code {
-               // No slot is ever written, so the walker must not
-               // pre-size a nesting map this grammar leaves empty.
-               const SEEDS_NESTING: bool = false;
-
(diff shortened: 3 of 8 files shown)


big-code-analysis-book/src/recipes/ci.md
```diff
@@ -868,7 +868,12 @@ Applies regardless of provider:
   cgroup-/cpuset-/quota-aware on Linux, OS CPU count on
   macOS/Windows — so CI runners no longer need to thread
   `--jobs "$(nproc)"` through every recipe. `--jobs 1`
-  remains a debugging knob, not a default.
+  remains a debugging knob, not a default. It is *not*
+  memory-aware: each worker holds one whole file in memory at a
+  time, and a parse-only run measures 25–70× the source size, so
+  peak RSS is roughly `jobs × 70 × the largest source file` in
```
dekobon: **Location:** `big-code-analysis-book/src/recipes/ci.md:874`

**MEDIUM -- Scope the RSS sizing rule to output modes that do not retain completed results**

The new formula is presented as generic CI guidance, but it is not a process-wide bound for every documented `metrics` destination. `metrics --output <FILE>` deliberately sends every `FuncSpace` over an unbounded channel and only collects/writes them after the walk (`big-code-analysis-cli/src/commands/analyze.rs:159-184`), so peak memory in that mode also grows with the full result set. Structured stdout has a similar extra term: `OrderedStdout` documents that its reorder buffer can degrade toward the whole tree when an early file dominates (`big-code-analysis-cli/src/ordered_stdout.rs:28-38`). Please qualify this as the parse/in-flight term (or scope it to `--output-dir`/`check`) rather than telling memory-limited CI users that lowering `--jobs` alone bounds RSS.

dekobon (commented):
## Code Review: perf(metrics): bound the cognitive nesting map

**Verdict: APPROVE WITH COMMENTS**

| Severity | Count |
|----------|-------|
| BLOCKER  | 0 |
| CRITICAL | 0 |
| HIGH     | 0 |
| MEDIUM   | 1 |

### Findings

| 1 | MEDIUM | `big-code-analysis-book/src/recipes/ci.md:874` | Scope the RSS sizing rule to output modes that do not retain completed results | correctness |

### Review passes executed

- Correctness and logic bugs
- Security review of the changed traversal state lifecycle
- Performance and scalability
- Test coverage

Files reviewed: 8

I did not find a metric-value or traversal-correctness regression in the slot reclamation itself. The only comment is on the new generic memory-sizing guidance, which omits existing output-buffer terms.

big-code-analysis-book/src/recipes/ci.md
dekobon: Fixed in `6424a983`. Verified both terms before editing, and both hold:

- `metrics --output <FILE>` builds a `crossbeam::channel::unbounded()` and only drains it with `rx.into_iter().collect::<Vec<_>>()` *after* `run_walk` returns (`commands/analyze.rs:161-183`), so the whole result set is live at once — the same shape as the #1116 regression, deliberate here.
- `OrderedStdout` documents the degradation itself (`ordered_stdout.rs:32-34`): the buffer approaches the whole tree "when one early file dominates the run". Worth noting that is precisely the tree this PR measures — a 28 MB `parser.c` among 100k small files — so it is the live case, not a hypothetical one.

The bullet now states the formula as the *in-flight* term and says `--jobs` bounds that term only, then splits the destinations: `--output-dir` writes per document and `bca check` reduces each file to its violations, so both stay at the in-flight bound; `--output <FILE>` and structured stdout add a result-set term. It closes by pointing a memory-limited runner at `--output-dir`.

The CHANGELOG sentence that referred to this rule was making the same overreach by reference, so it now says "the in-flight term in its `--jobs` sizing guidance" and names the two destinations that add to it.

`make pre-commit` passes.

Issue #1303, opened 2026-08-09
fix(cli): metrics/ops stdout document order is nondeterministic at --jobs > 1


dekobon:
## Summary

`bca metrics -O json` / `bca ops -O json` **to stdout** emit their
per-file documents in worker-completion order, so redirecting stdout to a
file gives a different byte sequence on every run at `--jobs > 1`. This
is the sibling of #1244 (which fixed the `--output <FILE>` aggregate) on
the destination people reach for most often in a pipeline.

## Evidence

40 trivial Python files, six identical invocations, measured after the
#1244 fix landed:

```console
$ for r in 1 2 3 4 5 6; do bca metrics --no-config -p . -O json -j 8 > stdout$r.json; done
$ # sha256 of each
stdout -O json distinct hashes over 6 runs: 6
```

`--output <FILE>` over the same tree: 1 distinct hash over 12 runs.
`--output-dir <DIR>`: identical trees (each document is its own file, so
order never enters).

## Why it is not simply the #1244 fix again

`--output` buffers every result before serializing, so sorting there is
free. Stdout *streams*: each worker writes its document as it finishes.
`walk_directory_seed` sorts the resolved file list, and its comment
already qualifies the guarantee as holding at `--jobs 1` only — so this
is known-by-construction rather than an oversight. Fixing it means
choosing one of:

- buffer the whole run and sort before writing (loses streaming, and the
  memory profile `--output` already has);
- emit in walk order by having each worker wait its turn (a reorder
  buffer keyed on the resolved file list's index);
- document the limitation and point users at `--output <FILE>` when they
  need reproducibility.

The third is cheap and may be the right answer; the second is the only
one that keeps both properties.

## Impact

Anyone piping `bca metrics -O json` into a file for diffing,
snapshotting, or CI caching — the same consumers #1244 was filed for.
`bca diff` is unaffected (it keys by `name`).

---

## Resolution Plan

Confirmed structurally on `main` (`d65b3049`). The stdout path is
per-document: `write_buffered(None, …)`
(`big-code-analysis-cli/src/formats.rs:251`) takes
`std::io::stdout().lock()` for the whole of one document, so two files'
output cannot interleave — but nothing orders the documents themselves,
and each worker writes as it finishes. That is the difference from
#1244, whose fix (`8bb990b1`, on the unmerged `batch-2026-08-08-2`
branch) could sort because `write_aggregate` already had every item in
hand.

1. **Choose between the three options first; recommend the reorder
[...]

Issue #1158, opened 2026-08-01
perf(cli): stream walked paths into the worker pool instead of materializing them


dekobon:
## Summary

Deferred half of #1114. That issue asked to feed the worker pool directly
from the directory walker over a bounded channel, "delet[ing] the
intermediate `Vec` for streaming consumers", so analysis of the first file
overlaps discovery of the last. #1114 shipped the parallel walk, the
opt-out `is_file()` re-stat, and moving dispatch off a reserved pool slot,
but **not** the streaming rearchitecture. This tracks it.

## Why it was deferred

Three reasons, recorded here so the next reader does not re-derive them.

**1. It conflicts with #1114's own determinism requirement.** Step 5 of that
issue asked to confirm output ordering does not depend on walk arrival
order. The fix shipped satisfies it by sorting the resolved list — which
also made per-file output order independent of readdir order, something the
old single-threaded walk never guaranteed. You cannot sort a stream. So
streaming reintroduces exactly the nondeterminism #1114 asked to rule out,
and the trade has to be made deliberately:

- keep sorting → no streaming;
- stream → per-file output order varies run to run at `--jobs 1`;
- stream *and* re-sort downstream → back to materializing, at the output
  boundary instead of the input one.

The third option may be the answer — most consumers of ordered output are
reading a rendered report, not the dispatch order — but it is a design
decision, not a mechanical change.

**2. `FilesData`'s shape is published API.** `FilesData` is `pub`,
re-exported from `lib.rs`, not `#[non_exhaustive]`, and constructed with
struct literals both in-tree and (potentially) downstream:

```rust
pub struct FilesData {
    pub paths: Vec<PathBuf>,
}

impl<Config> ConcurrentRunner<Config> {
    pub fn run(self, config: Config, files_data: FilesData) -> Result<(), ConcurrentErrors>;
}
```

Changing `run` to take a stream, or adding a field to `FilesData`, is a
SemVer break. The additive route is a second entry point — something like
`run_iter(config, impl IntoIterator<Item = PathBuf> + Send)` — with `run`
kept as a thin wrapper. Worth designing alongside whatever else lands in
the next major rather than bolting on now.

Note that #1114 already moved dispatch onto the calling thread, which makes
this much easier: `explore` no longer needs to live in a spawned producer,
so an iterator argument threads straight through.

**3. The measured headroom is small.** Warm, over
`tests/repositories/DeepSpeech` (12,732 files) on 16 cores:

[...]

Issue #1122, opened 2026-07-30
test(perf): Ancestors::checked re-derives every parent, making debug-build walks quadratic


dekobon:
## Summary

`Ancestors::checked` (`src/node.rs:482-490`) verifies the threaded ancestor chain with

```rust
debug_assert_eq!(
    chain.last().map(Node::id),
    node.parent().map(|parent| parent.id()),   // O(depth) tree-sitter FFI
    ...
);
```

Its own doc comment says `Node::parent` "is the O(depth) lookup #1084 exists to remove and must never run in a release build" — but under `cargo test` (debug) it runs **per node, on all five walks** that construct checked chains: the metric walk (`src/spaces/compute.rs:645`), ops (`src/ops.rs:276`), comment removal (`src/comment_rm.rs:103`), suppression scan (`src/suppression.rs:598`), and `act_on_node` (`src/node.rs:722`). Every test-build walk is therefore O(nodes x depth) even though the shipped walk is O(nodes) — `src/metrics/cognitive.rs` (~line 9795) already documents this and halved a test's nesting depth to live with it.

The tax lands on every parse in the suite (~2 314 unit-test parses + 1 437 corpus files) and is worst on the deep-nesting regression tests (`cognitive.rs` depths 1 000-2 000, `tokens.rs` depth 2 000, and the depth-2 000 wire/ops test measured at 11.6 s — see #1110).

The invariant is worth keeping — it caught real desyncs during #1084/#1088/#1096 — but not at per-node cost in every developer's inner loop.

## Resolution Plan

1. Move the check behind a dedicated opt-in cfg (e.g. `feature = "chain-audit"` or `cfg(chain_audit)`), off by default in dev/test.
2. Run one CI lane with it enabled (the existing all-features test job, or a dedicated small job over the corpus tests) so the invariant still has teeth; the dedicated chain-parity unit tests (`src/node.rs`, `src/getter.rs`, `src/checker.rs`, `src/function_tests.rs`) stay unconditional.
3. Alternatively (or additionally) sample: check when `chain.len()` is small or every Nth node — decide in review; the cfg gate is the simpler contract.
4. Confirm the deep-nesting tests' wall time collapses and that their depth comments (`cognitive.rs` ~9795) are updated to reflect the new cost model.
5. Cross-check with `docs/development/benchmarking.md`'s note that the unoptimised walk is quadratic — update that sentence once it no longer is.

## Assessment

| Dimension  | Rating |
|------------|--------|
[...]

PR #1378, opened 2026-08-30
docs: record lessons 91 and 92 and two rules

dekobon:
Two lessons and two rules from the 2026-08-29/30 window. Both lessons
name a mechanism nothing in the file covers; both non-qualifying
takeaways were routed to rules instead, per the file's own "standing
obligations live in rules, not here" preamble.

## Lesson 91 — a gate can filter out its own subject before the check runs

The belt-and-braces guard for RUSTSEC-2026-0258 would be a `[bans] deny`
on `h2 <0.4.16`. It is dead by construction: krates, cargo-deny's graph
builder, filters the crate out before the advisory and ban checks run
(`-L debug` logs `filtered h2 0.3.27`), on 0.19 and on the 0.20 that
CI's action bundles. Re-enabling `actix-web/http2` under both versions
left cargo-deny silent — so the entry would have looked like coverage
and could never have fired.

What makes it worth an entry is the shape: a config entry naming the
crate and the version range reads as a guard to every future
maintainer, and the tool exits 0, which is exactly what it prints when
the tree is genuinely clean. The working guard reads `Cargo.lock` from a
test, and `deny.toml` carries a comment where the dead entry would have
gone.

Cited by commit hash rather than PR — both are direct pushes verified
reachable on `main` with no PR, which is the one case the skill's
guardrail permits.

## Lesson 92 — an optimization's rationale can encode the waste it optimizes for

#1069 measured that the cognitive nesting map "converges on one entry
per visited node" and reserved `descendant_count()` up front to skip the
doubling chain. That is a genuine speedup — for growth that should not
have existed. The slot is dead after its children are seeded and was
never removed, so the reserve was sized to the leak: 540 MB of a
1,265 MB peak on a 28 MB generated `parser.c`, fixed in #1375 / PR
#1377.

The second mechanism is how it was found. All three hypotheses in the
issue (reorder buffer, path list materialized twice, glibc arenas) were
wrong, and code reading could not discriminate them because every
allocation suspected really is present. Bisecting by selection settled
it: `dump` measured 722 MB and every non-cognitive metric 723–725 MB
against cognitive's 1,264 MB, `--output-dir` matched stdout in peak, and
`bca check` — which has no reorder buffer at all — still peaked at
4.5 GB.

## Routed to rules instead

- **`AGENTS.md`**, GitHub workflow: re-verify a deferred issue's
  premises before building on them. #1158 proposed an output-ordering
  contract that #1303 had shipped the opposite of, with tests, and its
  own go/no-go measurement had never been run; running it closed the
  issue. This is the issue-tracker form of lesson 84.
- **`.claude/rules/testing.md`**: assert a whole-run invariant in the
  run, not in a fixture list. A `debug_assert!` at the end of a hot path
  recruits the entire corpus as input — 3,590 lib tests plus every
  integration suite, against the four fixtures a named test lists — and
  that is what established #1375's invariant across all twenty
  languages. Both guards, not either, and keep the assertion O(1): a
  per-node one is how the exact ancestor-chain check went quadratic
  (#1122).

## Notes

- Entries are 34 and 38 lines, inside the 45-line single-mechanism and
  ~60-line two-mechanism budgets. `rumdl` passes; both index anchors
  resolve.
- Numbering appends at 91/92, so no existing citation moves.
- Not done, and flagged rather than acted on: a one-clause merge of the
  `parser_cache` doc-claim instance into lesson 84 (one more instance of
  a mechanism it already documents with eight), and lesson 59's 81-line
  overrun against the 75-line stop.

### DIFF
--- a/docs/development/lessons_learned.md
+++ b/docs/development/lessons_learned.md
@@ -134,6 +134,8 @@ number and the higher number stays as a redirect.
 | [88](#88-a-text-scan-that-does-not-lex-the-language-measures-noise) | A text scan that does not lex the language measures noise |
 | [89](#89-a-positive-enumeration-and-a-negative-filter-differ-on-what-neither-names) | A positive enumeration and a negative filter differ on what neither names |
 | [90](#90-re-reading-a-single-consumption-source-yields-empty-not-an-error) | A re-read of a consumable source yields empty, not an error |
+| [91](#91-a-gate-can-filter-out-its-own-subject-before-the-check-runs) | A gate can filter out its own subject before the check runs |
+| [92](#92-an-optimizations-rationale-can-encode-the-waste-it-optimizes-for) | An optimization's rationale can encode the waste it optimizes for |
 
 ---
 
@@ -3542,5 +3544,80 @@ the project had exempted. The fix materializes the list once, before
 the walk, and carries it out in `CheckWalk::seeds`; the pre-existing
 regression test for #497 passed throughout, because it spelled the
 seed list as a file.
+
+---
+
+## 91. A gate can filter out its own subject before the check runs
+
+**Lesson:** When you add a dependency or supply-chain gate, prove it
+fails by reintroducing the condition it guards and watching it go red.
+A tool that builds a filtered view of its input before applying your
+rules can drop the very artifact a rule names, leaving an entry that is
+syntactically valid, semantically correct, and permanently silent. When
+the tool's view is derived, guard the raw artifact instead — the
+lockfile, the manifest — and record at the site where the dead entry
+would go why it is not there.
+
+A config entry reads as coverage. A future maintainer greps `deny.toml`,
+finds the crate and the version range spelled out, and stops looking;
+the tool exits 0, which is the same output it gives when the tree is
+genuinely clean. Nothing distinguishes "no violation" from "the subject
+was never in scope", and the gap opens precisely when the guard matters,
+because the condition that reintroduces the vulnerable crate is also the
+condition that puts it back beyond the filter.
+
+**cargo-deny cannot guard `h2 0.3.27`** (RUSTSEC-2026-0258, Scorecard
+alert #776; `3df3b2c3`, `b60943b5` — direct pushes to `main`, no PR).
+The crate reached the tree only through `actix-web`'s default `http2`
+feature, and no patched 0.3 release exists, so the fix was to drop the
+feature: `bca-web` binds plaintext and actix negotiates HTTP/2 only over
+TLS ALPN. The obvious belt-and-braces guard, a `[bans] deny` on
+`h2 <0.4.16`, turned out to be dead by construction — krates, cargo-deny's
+graph builder, filters the crate out before the advisory and ban checks
+run (`-L debug` logs `filtered h2 0.3.27`), on 0.19 and on the 0.20 that
+CI's action bundles. Re-enabling `actix-web/http2` under both versions
+left cargo-deny silent. The working guard reads `Cargo.lock` directly
+from a test, and `deny.toml` carries a comment where the entry would
+have been, naming the reproduction so the next reader does not add one.
+
+---
+
+## 92. An optimization's rationale can encode the waste it optimizes for
+
+**Lesson:** When a collection is pre-sized to its input, ask whether it
+needs to reach that size before optimizing for it — a reserve is
+evidence that somebody measured the final size, never evidence the
+growth was necessary. And attribute a resource peak by bisecting the
+run: one metric at a time, one output destination at a time, one worker
+count at a time. Reading the code ranks hypotheses it cannot
+discriminate, because every allocation you suspect really is there.
+
+A leak stated as a steady state stops looking like a leak. The sentence
+justifying the reserve asserts the growth as a fact, so review checks
+the arithmetic — is `descendant_count` the right size? — rather than the
+premise, and the premise is the bug. Nothing else flags it either: the
+retained entries are dead, so no metric value, snapshot, or exit status
+moves whether they are freed or kept.
+
+**The nesting map's reserve was sized to the leak it should have
+prevented** (#1069, fixed in #1375 / PR #1377). `NestingMap` carries one
+`Nesting` per node id, seeded by the parent, read once by the node's own
+`Cognitive::compute`, read once more when its children are seeded — and
+then never removed. #1069 measured that it "converges on one entry per
+visited node", and reserved `descendant_count()` up front to skip the
+doubling chain, which is a real speedup for a growth that should not
+have existed. On a 28 MB generated tree-sitter `parser.c` that was
+540 MB of a 1,265 MB peak. Freeing each slot in its last reader leaves
+the live set bounded by the traversal stack: 735 MB, and 3.3 s against
+5.1 s.
+
+**All three hypotheses in the issue were wrong** (#1375). It proposed
+the reorder buffer, a path list materialized twice, and glibc arena
+retention. Selection settled it in minutes where code reading had not:
+`dump` (parse only) measured 722 MB and every non-cognitive metric
+723–725 MB against cognitive's 1,264 MB, `--output-dir` matched stdout
+byte-for-byte in peak — ruling out both the wire clone and the reorder
+buffer — and `bca check`, which has no reorder buffer at all, still
+peaked at 4.5 GB on the same tree.
 
 ---
--- a/docs/development/lessons_learned.md
+++ b/docs/development/lessons_learned.md
@@ -3570,8 +3570,12 @@ condition that puts it back beyond the filter.
 alert #776; `3df3b2c3`, `b60943b5` — direct pushes to `main`, no PR).
 The crate reached the tree only through `actix-web`'s default `http2`
 feature, and no patched 0.3 release exists, so the fix was to drop the
-feature: `bca-web` binds plaintext and actix negotiates HTTP/2 only over
-TLS ALPN. The obvious belt-and-braces guard, a `[bans] deny` on
+feature: this daemon reaches the network through a plain
+`HttpServer::bind`, which serves HTTP/1.x only. Note the reason is that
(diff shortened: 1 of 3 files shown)


docs/development/lessons_learned.md
```diff
@@ -3544,3 +3546,78 @@ regression test for #497 passed throughout, because it spelled the
...
+
+A leak stated as a steady state stops looking like a leak. The sentence
+justifying the reserve asserts the growth as a fact, so review checks
+the arithmetic — is `descendant_count` the right size? — rather than the
+premise, and the premise is the bug. Nothing else flags it either: the
+retained entries are dead, so no metric value, snapshot, or exit status
+moves whether they are freed or kept.
+
+**The nesting map's reserve was sized to the leak it should have
+prevented** (#1069, fixed in #1375 / PR #1377). `NestingMap` carries one
```
dekobon, line 3607: **Location:** `docs/development/lessons_learned.md:3603`

**HIGH -- This lesson currently lands before the fix it describes**

PR #1377 is still open, and #1378 is based directly on `main` rather than on `perf/1375-nesting-map-lifetime`. In this branch `src/spaces/compute.rs` still allocates `NestingMap::with_capacity_and_hasher(node.descendant_count(), ...)`, still uses `get` in `propagate_nesting_to_children`, and has no `nesting_slots_retained` or final `debug_assert!`. If this merges first, both this lesson and the new testing rule state that the leak is fixed while `main` still contains the old behavior. Please stack/rebase this branch on #1377 or ensure #1377 lands first.

docs/development/lessons_learned.md
```diff
@@ -3544,3 +3546,78 @@ regression test for #497 passed throughout, because it spelled the
...
+genuinely clean. Nothing distinguishes "no violation" from "the subject
+was never in scope", and the gap opens precisely when the guard matters,
+because the condition that reintroduces the vulnerable crate is also the
+condition that puts it back beyond the filter.
+
+**cargo-deny cannot guard `h2 0.3.27`** (RUSTSEC-2026-0258, Scorecard
+alert #776; `3df3b2c3`, `b60943b5` — direct pushes to `main`, no PR).
+The crate reached the tree only through `actix-web`'s default `http2`
+feature, and no patched 0.3 release exists, so the fix was to drop the
+feature: `bca-web` binds plaintext and actix negotiates HTTP/2 only over
```
dekobon: **Location:** `docs/development/lessons_learned.md:3573`

**MEDIUM -- Actix also supports plaintext HTTP/2 via H2C**

`HttpServer` 4.15 exposes `bind_auto_h2c()` and `listen_auto_h2c()` for plaintext HTTP/2; only `bca-web`'s current `.bind((host, port))` call keeps HTTP/2 unreachable here. Saying Actix negotiates HTTP/2 only over TLS ALPN makes the rationale false and would mislead a future switch to H2C. Please tie the statement to this daemon using `.bind()` rather than to Actix generally.

docs/development/lessons_learned.md
```diff
@@ -3544,3 +3546,78 @@ regression test for #497 passed throughout, because it spelled the
...
+A leak stated as a steady state stops looking like a leak. The sentence
+justifying the reserve asserts the growth as a fact, so review checks
+the arithmetic — is `descendant_count` the right size? — rather than the
+premise, and the premise is the bug. Nothing else flags it either: the
+retained entries are dead, so no metric value, snapshot, or exit status
+moves whether they are freed or kept.
+
+**The nesting map's reserve was sized to the leak it should have
+prevented** (#1069, fixed in #1375 / PR #1377). `NestingMap` carries one
+`Nesting` per node id, seeded by the parent, read once by the node's own
```
dekobon: **Location:** `docs/development/lessons_learned.md:3604`

**MEDIUM -- The slot lifecycle omits the paths that needed separate freeing**

Before #1377, a leaf slot was not read again during child seeding because `propagate_nesting_to_children` returned before looking it up when `children.is_empty()`, and an `exclude_tests`-pruned node skipped both `compute` and propagation. Those are exactly why #1377 removes the leaf bail-out and adds a separate prune-arm removal. As written, this says every slot had the same two reads and hides the edge cases the fix has to preserve.

docs/development/lessons_learned.md
```diff
@@ -3544,3 +3546,78 @@ regression test for #497 passed throughout, because it spelled the
...
+`Cognitive::compute`, read once more when its children are seeded — and
+then never removed. #1069 measured that it "converges on one entry per
+visited node", and reserved `descendant_count()` up front to skip the
+doubling chain, which is a real speedup for a growth that should not
+have existed. On a 28 MB generated tree-sitter `parser.c` that was
+540 MB of a 1,265 MB peak. Freeing each slot in its last reader leaves
+the live set bounded by the traversal stack: 735 MB, and 3.3 s against
+5.1 s.
+
+**All three hypotheses in the issue were wrong** (#1375). It proposed
```
dekobon: **Location:** `docs/development/lessons_learned.md:3614`

**MEDIUM -- This attribution paragraph sharpens the evidence past what #1375 established**

#1375's body does not propose the twice-materialized path list as one of its three suspects; that came from #1158. Its follow-up says the body suspects were secondary, not all wrong, and specifically records `MALLOC_ARENA_MAX=1` lowering the peak to 3.4 GB, so glibc arena retention was real even if it was not the dominant term. The same table reports stdout at 1,265 MB and `--output-dir` at 1,266 MB, not "byte-for-byte" identical. Please preserve those measured distinctions.

dekobon (commented):
## Code Review: docs: record lessons 91 and 92 and two rules

**Verdict: REQUEST CHANGES**

| Severity | Count |
|----------|-------|
| HIGH | 1 |
| MEDIUM | 3 |

### Findings

| 1 | HIGH | `docs/development/lessons_learned.md:3603` | This lesson currently lands before the fix it describes | correctness |
| 2 | MEDIUM | `docs/development/lessons_learned.md:3573` | Actix also supports plaintext HTTP/2 via H2C | correctness |
| 3 | MEDIUM | `docs/development/lessons_learned.md:3604` | The slot lifecycle omits the paths that needed separate freeing | correctness |
| 4 | MEDIUM | `docs/development/lessons_learned.md:3614` | This attribution paragraph sharpens the evidence past what #1375 established | correctness |

### Review passes executed

- Documentation correctness and logic
- Factual verification against issue, PR, and source history
- Documentation consistency and test-guidance review

Files reviewed: 3 changed files, plus the referenced source and issue history.

docs/development/lessons_learned.md
dekobon: Fixed — you are right, and my split reasoning was wrong. I branched off `main` to keep the unrelated h2 lesson out of the code fix, but that traded a topical problem for a correctness one: merged first, both the lesson and the testing rule would describe a fix `main` does not contain.

Rebased onto `perf/1375-nesting-map-lifetime` and retargeted this PR's base to it, so #1378 is now stacked on #1377 and its diff is just the two docs commits. It cannot land first.

docs/development/lessons_learned.md
dekobon: Fixed, and thank you — the claim was false and I inherited it from `3df3b2c3`'s own commit message rather than checking it.

Verified: `bind_auto_h2c` is at `actix-web-4.14.0/src/server.rs:513` (`listen_auto_h2c` at 685), so actix does speak plaintext HTTP/2. The entry now ties the rationale to this daemon's plain `HttpServer::bind` at `web/server.rs:303` instead.

One thing the check turned up that strengthens the fix: `bind_auto_h2c` is itself `#[cfg(feature = "http2")]`, so dropping the feature removes the H2C entry point along with the vulnerable crate. That is now in the entry, since it means the guard does not rest on the bind call staying as it is.

docs/development/lessons_learned.md
dekobon: Fixed. The simplification hid exactly the two shapes the fix has to treat specially, which is the opposite of what the entry should leave a reader with.

It now names both: a leaf returned early on `children.is_empty()`, and an `exclude_tests`-pruned node reached neither `compute` nor propagation — so freeing in the last reader means dropping the leaf bail-out and adding a removal on the prune arm.

docs/development/lessons_learned.md
dekobon: Fixed, all three. Checked each against the issue body and the follow-up:

- The twice-materialized path list was #1158's suspect, not #1375's. #1375 listed the per-file working set, the reorder buffer, and arena retention. Corrected.
- "All three wrong" overstated it. `MALLOC_ARENA_MAX=1` did reach 3.4 GB, so arena retention was real. The heading is now "every suspect the issue named turned out to be secondary", and the paragraph says none was wrong to suspect and the last was demonstrably real.
- `--output-dir` was 1,266 MB against stdout's 1,265 MB. "Byte-for-byte" is now "within a megabyte", with both numbers.

Worth noting this finding is lesson 84 catching the entry that sits two above it in the same file.

dekobon:
Landed on `main` — GitHub shows this as closed rather than merged, which is an artifact of how it went in, not of the work being dropped.

`main` had not moved from the branch point, and the repo ruleset `protect-main` requires `required_signatures` **and** `required_linear_history`, so neither merge button was usable: GitHub refuses a rebase merge outright ("Rebase merges cannot be automatically signed by GitHub") and a merge commit violates linear history. A fast-forward satisfies every rule and keeps the original GPG signatures, so the stack went in that way:

```
9c93e053 sig=G docs(lessons): correct four review findings in 91 and 92  <- this PR
632345d8 sig=G docs: record lessons 91 and 92 and two rules              <- this PR
6424a983 sig=G docs(ci): scope the RSS rule to the in-flight term        <- #1377
03cb037f sig=G perf(metrics): bound the cognitive nesting map            <- #1377
```

#1377 registered as merged because its head was reachable from its base. This PR could not: its base was the stacked branch, and once `main` and `docs/lessons-91-92` became identical GitHub refused to retarget it ("There are no new commits between base branch main and head branch"). Deleting the merged parent branch then closed it.

Both commits from this PR are on `main` and signed. All four review findings were addressed in `9c93e053`.

Issue #1426, opened 2026-09-12
test(features): single-language legs fail to build tests/api/ast_seam_test.rs


dekobon:
## Summary

**Fixed** in `test(features): build tests/api under single-language feature
sets`.

Two single-language feature subsets were reported failing on `main`, both in
`tests/api/ast_seam_test.rs` rather than in the library — but at *different*
stages, which matters for where a gate would have to sit:

```console
$ cargo check --no-default-features --features c -p big-code-analysis --all-targets
error[E0433]: cannot find type `PathBuf` in this scope
error[E0433]: cannot find type `MetricsOptions` in this scope
error: could not compile `big-code-analysis` (test "api") due to 2 previous errors

$ cargo clippy --no-default-features --features tcl -p big-code-analysis --all-targets -- -D warnings
error: for loop over a single element
error: could not compile `big-code-analysis` (test "api") due to 1 previous error
```

- `--features c` is a genuine **rustc** failure: unresolved `PathBuf` /
  `MetricsOptions` imports, reachable only once the `rust`-gated block is
  compiled out.
- `--features tcl` **compiles clean** (`cargo check` passes) and fails only
  under `clippy -D warnings`: `clippy::single_element_loop`, from a loop whose
  row table reduces to one element when the other languages are gated out.

Both reproduced directly on the current tree, not inferred.

Found while fixing #1286 / #1411. Verified pre-existing by stashing that work
and rebuilding, so it is untouched by that change and predates this batch.

## What the full sweep found

Sweeping **all 22** single-language subsets — not just the two above — under
`cargo clippy --no-default-features --features <lang> -p big-code-analysis
--all-targets -- -D warnings` gave **5 failing legs from 3 distinct defects**:

| Leg(s) | Defect | Diagnostic |
|---|---|---|
| `c` | `PathBuf` / `MetricsOptions` unions too **narrow** — gated `any(rust, python, cpp)`, used by a `c`-gated test | `E0433` |
| `cpp`, `python` | `ops_tree_contains_operator` union too **wide** — gated `any(rust, python, cpp)`, one `rust`-gated caller | `dead_code` |
| `tcl`, `irules` | two-row Tcl-family table with cfg-gated rows collapses to a singleton | `clippy::single_element_loop` |

The `cpp` / `python` class was not in the original report. It is the exact
mirror of the `c` class: the same hand-maintained union, wrong in the other
direction.

`big-code-analysis-ast` is clean on all 22 single-language legs, before and
after.

[...]

Issue #1415, opened 2026-09-11
test(grammars): alterator_string_flattening rows panic under a partial feature set


dekobon:
## Summary

`tests/grammars/alterator_string_flattening.rs` builds its case table
with `flatten_cases!`, which emits a bare `#[test]` per row, and
`tests/grammars/main.rs` declares `mod alterator_string_flattening;`
ungated. Every row calls `dump_root`, whose

```rust
.expect("language feature enabled under --all-features")
```

panics when the row's grammar feature is off. So the whole file fails
under any feature set that does not enable all eight of its languages.

This is the shape `.claude/rules/testing.md` names under "Gate a
feature-gated fixture table on the union of its rows", and the
false-failure history it cites (#1220, PR #1221) is exactly what it
produces: a build that reads as a defect in whatever was being changed.

## Evidence

```text
$ cargo test -p big-code-analysis --no-default-features --features go \
      --test grammars --no-fail-fast
test result: FAILED. 0 passed; 11 failed
  csharp_flattens_string_literal, elixir_flattens_string_literal,
  irules_flattens_braced_value, irules_flattens_quoted_word,
  lua_flattens_string_literal, mozcpp_flattens_string_literal,
  objc_flattens_string_literal, ruby_flattens_string_literal,
  tcl_flattens_braced_value, tcl_flattens_quoted_word,
  tcl_flattens_quoted_word_inside_a_proc_body
```

## Why it is invisible today

The `feature-matrix` job in `.github/workflows/ci.yml` runs
`cargo check --all-targets`, never `cargo test`, and every other leg
runs `--all-features`. Nothing in CI executes this file under a partial
feature set, so the gap only surfaces for a contributor building a
subset locally — or the first time someone adds a `cargo test` leg to
the matrix.

## Suggested fix

Give `flatten_cases!` a per-row feature name and gate each generated
`fn`, so a row is *absent* rather than failing:

```rust
macro_rules! flatten_cases {
    ($($name:ident: $feat:literal, $lang:expr, $code:expr, $file:expr, $lit:expr;)*) => {
        $(
            #[test]
            #[cfg(feature = $feat)]
            fn $name() { assert_flattened($lang, $code, $file, $lit); }
        )*
    };
}
```

Note the union is over *features*, not languages: `LANG::Tsx` rides
`feature = "typescript"`, and `LANG::Mozcpp` rides `feature = "cpp"` —
check each row against `Cargo.toml` rather than assuming the slug.

[...]

Issue #1413, opened 2026-09-11
test: per-language tests naming a concrete *Parser are ungated and panic under a single-language build


dekobon:
## Summary

Roughly 158 PHP-named tests in `src/metrics/loc.rs` (and the same shape in
sibling metric modules) call `check_metrics::<PhpParser>` with no
`#[cfg(feature = "php")]` gate on the `fn`. The parser *type alias* is defined
unconditionally by `mk_langs!`, so these compile in any feature set — but
`LANG::get_ts_language` is cfg'd, so under a build without the language every
one of them panics:

```
invariant: the caller checked LANG::is_enabled … LanguageDisabled(Php)
  at big-code-analysis-ast/src/node.rs:84
```

Reproduce with `cargo test -p big-code-analysis --no-default-features --features rust`.

This is not PHP-specific — it is the general shape wherever a per-language test
names a concrete `*Parser` without gating on that language's feature. PHP is
simply where it was measured, while fixing #1396.

## Why it has not been noticed

`make pre-commit` clippies and tests only the default-features and
`--all-features` flavours, both of which enable every language. The
feature-matrix CI job runs `cargo check --all-targets`, which compiles the
tests but does not run them, so a panicking-but-compiling test is invisible
there too. Nothing in the gate executes a single-language build's test suite.

## What correct looks like

`.claude/rules/testing.md` already states the rule for fixture *tables*
(`#[cfg(any(feature = …))]` on the `fn` plus a non-vacuity guard). The
single-language case is the simpler half: `#[cfg(feature = "<lang>")]` on the
`fn`, so the test is **absent** rather than panicking.

The four #1396 tests added in `src/metrics/loc.rs` carry that gate and are
correctly absent under `--features rust`; they are the intended pattern, and
their ungated neighbours are the defect.

## Scope

Worth measuring before committing to a fix — the count above is for
`src/metrics/loc.rs` alone, and the other twelve metric modules plus
`src/spaces_tests.rs` are likely in the same state. A mechanical pass adding
the gate is straightforward; the judgement call is whether to also add a
feature-matrix CI leg that actually *runs* a single-language test suite, since
without one the gate cannot regress-test itself.

dekobon:
## Measured scope, and two corrections to this issue

Measured against `main` at `a3041252` while triaging this issue for a batch fix.
It was deferred rather than attempted: the scope is roughly an order of magnitude
larger than the body implies, and it does not fit alongside smaller fixes.

### The count

[...]

Issue #1411, opened 2026-09-10
test(abc): two cross-language ABC table modules fail on a feature subset instead of compiling out


dekobon:
## Summary

`.claude/rules/testing.md`, "Gate a feature-gated fixture table on the
union of its rows", requires **both** halves: a `checked > 0` /
`assert_fixtures_present` non-vacuity guard *and* a
`#[cfg(any(feature = …))]` union on the `fn`, so a build enabling none of
the table's languages leaves the test **absent** rather than failing.

Two cross-language table modules in `src/metrics/abc.rs` have the guard
and not the gate — the exact #1220 / PR #1221 false-failure shape the rule
was written from:

- `ternary_comment_invariance` (`src/metrics/abc.rs:11261`)
- `keyword_negation_parity` (`src/metrics/abc.rs:11390`)

## Reproduction

```console
$ cargo test -p big-code-analysis --lib --no-default-features --features go
```

Four tests fail, panicking at `abc.rs:11439` and `:11473` with "no
language enabled; this test asserted nothing". Nothing is wrong with the
code under test — `LANG::Go` simply has no row in either `cases()` table,
so `checked` stays 0 and the guard fires as a failure instead of the test
being compiled out.

The failure reads as a defect in whatever the contributor was changing,
which is precisely what the rule exists to prevent.

## Fix shape

Add the feature union to each module, naming the union of the features
its rows use (features, not languages — `LANG::Tsx` rides
`feature = "typescript"`, and `LANG::Mozcpp` / `LANG::Objc` ride their
own). Keep the `checked > 0` guard: it still covers the residual case
where a runtime `is_enabled()` check stops agreeing with the feature it
compiled under.

`numeric_bool_operands` in the same file (added in #1379, commit
`6a998d77`) has both halves and is the shape to copy.

## Verification

`--features go` is the canonical reproducer — `rust,typescript` is not a
valid check for either table, since both contain TypeScript rows. After
the fix, each module must report **0 tests matched** under
`--features go`, not 0 failures.

## Provenance

Found by an `audit-tests` pass over #1379's new tests; pre-existing and
untouched by that commit.

## Resolution

Fixed on `fix/issue-1286` (commit `057131cc`), in the same commit as
#1286.

**This issue is a strict subset of #1286.** Its two modules —
`ternary_comment_invariance` and `keyword_negation_parity`, both in
`src/metrics/abc.rs` — are two of that issue's twenty-plus sites, and
[...]

PR #1427, opened 2026-09-12
test(features): gate language-specific tests on their Cargo features

dekobon:
Fixes six issues, all but two variations on one theme: **a test that
exercises a language must be gated on that language's Cargo feature**, so
a partial-feature build leaves it *absent* rather than failing with panics
that read as a defect in whatever the contributor was changing — the
#1220 / PR #1221 shape `.claude/rules/testing.md` was written from.

| Commit | Issue | |
|---|---|---|
| `1591bbdc` | #1234 | Gate excluded-crate lockfile staleness |
| `d525e6df` | #1235 | `fuzz-smoke` reports every failing target |
| `bd3001b6` | #1415 | Gate the 24 `flatten_cases!` rows per row |
| `2f0bbd11` | #1281 | Parity fixtures on an exhaustive `match`, +107 rows |
| `7fba2483` | #1286, #1411 | Gate 25 sweeps, split one, add 4 missing guards |
| `f703ad02` | — | Whole-branch review remediation |

## What each one does

**#1234** — a workspace-excluded crate carries its own `Cargo.lock`, and
`cargo update --workspace` (which `RELEASING.md` calls mandatory during a
bump) reaches none of them. Every gate that consumes one passes
`--locked`, so a bump that skipped them turned `make enums-check` /
`fuzz-check` / `release-check` red on a later, unrelated commit.
`check-versions.py` now compares every path package against the version
that package's own manifest declares, and names the lockfile, the
package, both versions, and the `cargo update --manifest-path` line that
repairs it.

Candidates are derived from the root manifest's `[workspace] exclude`
array rather than from the gate's existing `EXCLUDED_LEAF_DIRS` tuple.
That tuple encodes the *first* invariant's admission rule ("must equal
the canonical workspace version"), which is exactly why `fuzz` —
deliberately `0.0.0` — is not in it; reusing it would have silently
dropped one of the two lockfiles the issue exists for.

**#1235** — `.SHELLFLAGS := -eu` meant the first crashing `cargo fuzz
run` ended the loop. `fuzz-smoke` is what the quarterly cron invokes, the
run nobody watches, so a quarter with three unrelated crashes reported
one and rediscovered the rest only after it was fixed. The loop now
collects per-target status and exits naming every failing target.
`fuzz-replay`, the per-PR gate, deliberately still stops at the first.

The workflow's advisory status is now recorded as a decision with its
rationale rather than stated as a bare fact.

**#1415** — every row of `alterator_string_flattening` panics when its
grammar is compiled out, and the module is declared ungated, so any
feature set short of all eight red-Xed up to 24 tests. `flatten_cases!`
now takes a per-row feature literal. A whole-module gate cannot express
this: the table spans eight features, which is why its single-grammar
siblings could use one and this file cannot.

**#1281** — five hand-listed parity suites now key fixtures on an
exhaustive `match` over `LANG`, so a new language cannot be added without
deciding whether it spells each construct. Adds 107 measured rows across
11 families with a reason at every `None` arm, replacing `nargs`'s false
"every supported language …" prose list and `nexits`'s asserted "all 23
languages" claim with a compiler-enforced one. None of the 107 moved a
metric; all were at parity when added.

It also fixes an unfiled defect in the same files: none of them filtered
on `is_enabled()`, so 13 of 19 parity tests panicked under
`--no-default-features --features go`. Now 16 of 16 pass.

**#1286 / #1411** — 25 sweeps had the loud non-vacuity guard without the
`#[cfg(any(feature = …))]` union that makes a test absent rather than
failing. 12 failed under `--features go` and 9 under `rust,typescript`;
both are now 0. Four more had the inverse problem — a filter and no guard,
so they passed having asserted nothing.

Two sites are deliberately not plain unions:

- `comments_in_parameter_lists` has a guard that is a **conjunction over
  two disjoint populations**, which no union can satisfy — a union admits
  every configuration satisfying *one* clause while the guard demands all
  of them. Split into two tests, each with its own gate and guard.
- `c_family_return_type_declarators` needed the opposite, an
  **intersection** (`any(cpp, mozcpp)`), because its `cpp_only` clause
  dominates the other two.

## Verification

`make pre-commit` → `BCA_GATE: pass` on every commit and on the final
branch. For each gating change: the test is absent (not merely passing)
under a subset enabling none of its features, present and passing under
one enabling a row, and the `--all-features` `cargo nextest list` set is
**diffed as a set** before and after. The only intended membership change
in the whole branch is the one-into-two split above.

The partial-feature suites do **not** go green overall, and are not meant
to: ~2,600 ungated per-language tests remain. That is #1413.

## Deferred and filed

- **#1285 / #1413 deferred** after measurement: 2,613 ungated tests naming
  a concrete `*Parser` in `src/` plus 54 in the ast crate, against exactly
  4 gated. #1285's plan counts on gating at `mod` granularity, but the
  per-language files under `src/metrics/<metric>/` are *implementation*
  modules holding 28 tests in total — every per-language test is inline in
  one `mod tests` per metric file, so that route does not exist. Both
  issues carry the full measurement.
- **#1426 filed** — `--no-default-features --features c` fails `cargo
  check` and `--features tcl` fails `clippy -D warnings`, both in
  `tests/api/ast_seam_test.rs`. Pre-existing, invisible to CI, and it
  breaks the exact reproducer subsets `.claude/rules/testing.md`
  prescribes.

## Issue claims that did not survive verification

Corrected in each issue's thread:

- #1234's plan would have missed `fuzz/Cargo.lock` (above).
- #1235 said `main`'s `ci` aggregator gates merges. `main` is governed by
  the `protect-main` **ruleset** — the legacy
  `/branches/main/protection` endpoint answers `404 Branch not protected`
  for a ruleset-governed branch, which is where that reading came from.
  The ruleset names no required status checks, so the advisory decision
  stands, but the premise was wrong and the doc now cites
  `gh api repos/:owner/:repo/rules/branches/main`.
- #1415 said 11 rows (24 — `21230d4a` added the rest) and gave
  `--features go` as the reproducer; `default = ["all-languages"]` makes
  that additive, so it proves nothing without `--no-default-features`.
  It also said `LANG::Mozcpp` rides `cpp`; it has its own feature.
- #1281 was stale on #1272, which shipped — Elixir rows already existed in
  four families. And `LANG::C` was absent only from the *hand-listed*
  suites, not from all of them.
- #1286's `src/c_declarator.rs` path moved to
  `src/c_family_space_names_tests.rs` in the #1376 split.
- #1411 misses that `keyword_negation_parity` holds a third test, which a
  module gate naming only the other two would have dropped from a
  Lua-only build.

## What the whole-branch review caught

`7fba2483` gated `container_scope_tests` on the 16 features its
`FIXTURES` table uses. A module gate is an **outer** `cfg`, so the one
test in that module reading a different table —
`a_language_with_no_member_construct_emits_neither_block`, gated
`any(bash, lua, c)` — stopped existing under exactly the builds it
covers. Silent, and invisible to CI, whose feature matrix enables none of
those three.

Widening the module gate alone would have traded that for the spurious
failure the gate exists to prevent, since `FIXTURES` is empty under
`bash`. `f703ad02` inverts it instead: the module names every feature any
of its tests uses, and the narrower `FIXTURES` union moves onto the four
sweeps that read `FIXTURES`.

Worth noting that this branch's own CHANGELOG describes catching the
identical trap in `abc`'s Lua-only third test. Same shape, two modules,
caught in one and not the other.

## Note for merging

Every commit is **unsigned** — GPG pinentry had no tty during the run.
`protect-main` enforces `required_signatures`, so either re-sign first:

```bash
git rebase --exec 'git commit --amend -S --no-edit' origin/main
```

(safe — the branch is linear, zero merge commits), or squash-merge, which
has GitHub create and sign the resulting commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1228,17 +1228,36 @@ fuzz-check:
 	  echo "nightly toolchain or cargo-fuzz not found; skipping fuzz-check"; \
 	fi
 
+# The loop collects failures instead of stopping at the first, because
+# this is what the quarterly cron runs and nobody is watching it. A hunt
+# that trips over three unrelated crashes should report three; stopping
+# at the first hides the other two until the quarter after the fix, and
+# the artifact upload in fuzz.yml already globs every target's
+# reproducer. `.SHELLFLAGS` carries `-e`, but a command on the left of
+# `||` is not a fatal failure under it, so the run survives a crashing
+# target and still exits non-zero at the end.
+#
+# `fuzz-replay` below keeps the opposite shape deliberately: it is the
+# per-PR gate, where one broken seed is already enough to stop the
+# merge, and the seconds saved by not replaying the rest are the point.
 fuzz-smoke:
 	@if rustup toolchain list 2>/dev/null | grep -q '^nightly' && \
 	    command -v cargo-fuzz >/dev/null 2>&1; then \
 	  $(FUZZ_REQUIRE_SYMBOLIZER); \
+	  failed=""; \
 	  for t in $$(cargo +nightly fuzz list); do \
 	    echo "Fuzzing $$t for $(FUZZ_RUNS) runs..."; \
 	    mkdir -p $(FUZZ_WORK)/"$$t"; \
 	    $(FUZZ_CC_ENV) $(FUZZ_RUN_ENV) cargo +nightly fuzz run --target $(FUZZ_HOST_TRIPLE) "$$t" \
 	      $(FUZZ_WORK)/"$$t" $(BASE_DIR)fuzz/corpus/"$$t" \
-	      -- -runs=$(FUZZ_RUNS) -timeout=$(FUZZ_TIMEOUT); \
+	      -- -runs=$(FUZZ_RUNS) -timeout=$(FUZZ_TIMEOUT) \
+	      || failed="$$failed $$t"; \
 	  done; \
+	  if [ -n "$$failed" ]; then \
+	    echo "fuzz-smoke: failing targets:$$failed"; \
+	    echo "reproduce with: make fuzz-run FUZZ_TARGET=<target> FUZZ_INPUT=fuzz/artifacts/<target>/<file>"; \
+	    exit 1; \
+	  fi; \
 	else \
 	  echo "nightly toolchain or cargo-fuzz not found; skipping fuzz-smoke"; \
 	fi
--- a/tests/parity/cyclomatic_cross_language_parity.rs
+++ b/tests/parity/cyclomatic_cross_language_parity.rs
@@ -23,15 +23,23 @@
 //! prescribes a cross-language parity test for this metric. That is
 //! what this file is.
 //!
-//! For every construct family below the test builds a
-//! `BTreeMap<&str, f64>` keyed by language with its standard-CCN sum, then
-//! asserts every value is equal **after subtracting a small set of
-//! explicitly-documented per-language offsets** that account for
-//! intrinsic, language-mandated structural differences (e.g. Java's
-//! requirement that every function live inside a class, which adds a
-//! `FuncSpace` and therefore +1 to every sum). On failure the full
-//! before/after maps are printed so any drift and the language that
-//! drifted are immediately visible.
+//! Each construct family below is an exhaustive `match` on [`LANG`]
+//! returning a [`Fixture`]: one arm per variant, so adding a language
+//! fails to compile until someone decides whether it spells the family's
+//! construct, and a language that does not says so with `None` and a
+//! reason at the arm. Before #1281 the families were hand-maintained
+//! `BTreeMap` inserts that had fallen behind the roster — Ruby, Lua,
+//! Perl and Go were absent from families they can express, and every row
+//! labelled `"c"` actually parsed [`LANG::Cpp`].
+//!
+//! [`check_family`] then builds a `BTreeMap<&str, f64>` keyed by
+//! language with its standard-CCN sum and asserts every value is equal
+//! **after subtracting a small set of explicitly-documented per-language
+//! offsets** that account for intrinsic, language-mandated structural
+//! differences (e.g. Java's requirement that every function live inside
+//! a class, which adds a `FuncSpace` and therefore +1 to every sum). On
+//! failure the full before/after maps are printed so any drift and the
+//! language that drifted are immediately visible.
 //!
 //! Each offset must be either (a) a structural language requirement
 //! that cannot be removed by rewriting the fixture, or (b) a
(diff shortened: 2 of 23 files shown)


dekobon (commented):
## Code Review: test(features): gate language-specific tests on their Cargo features

**Verdict: CHANGES NEEDED**

| Severity | Count |
|----------|-------|
| BLOCKER | 0 |
| CRITICAL | 0 |
| HIGH | 0 |
| MEDIUM | 3 |

### Findings

| 1 | MEDIUM | `utils/check-versions.py:447` | Missing checked-in lockfiles are skipped | correctness |
| 2 | MEDIUM | `tests/parity/cyclomatic_cross_language_parity.rs:180` | Parity tests pass after measuring zero languages | test-gap |
| 3 | MEDIUM | `Makefile:1247` | Failed/empty fuzz target discovery reports success | correctness |

### Review passes executed

- Correctness and logic bugs
- Rust feature-gate and parity-fixture review
- Test-quality and non-vacuity audit
- Release/version gate review
- Fuzz workflow and Makefile semantics review
- Security review plus Semgrep OSS scan

Files reviewed: 23. Reproductions were run against head `f703ad02137113e82a9dbe8c759f7411f877f6c2`; the PR worktree remained unmodified. GitHub CI is green, and `utils/check-versions-test.py` passes all 31 tests, but those checks do not cover the three false-success paths above.

utils/check-versions.py
```diff
@@ -272,6 +321,188 @@ def check_external_grammar_lockstep(root: pathlib.Path) -> list[str]:
...
+    An excluded crate need not have a `Cargo.lock` — one only exists
+    once something has resolved that crate's manifest — so a missing
+    lockfile is skipped rather than reported. A crate with no manifest
+    at all is a different matter and is rejected by
+    `owned_crate_dirs`.
+    """
+    return [
+        (crate, root / crate / "Cargo.lock")
+        for crate in _workspace_array(root, "exclude")
+        if crate not in NON_CRATE_EXCLUDES and (root / crate / "Cargo.lock").is_file()
```
dekobon, line 447: **Location:** `utils/check-versions.py:447`

**MEDIUM -- Treat a missing checked-in lockfile as drift**

Filtering candidates with `Cargo.lock.is_file()` makes deletion of one of the seven committed lockfiles a pass as long as any other lockfile remains. I reproduced this by deleting only `fuzz/Cargo.lock`: `python3 utils/check-versions.py` exited 0 and reported `6 excluded lockfiles in step`, while `cargo test --locked --manifest-path fuzz/Cargo.toml --no-run` failed with `cannot create the lock file ... because --locked was passed`. The all-or-nothing guard below does not catch a partial disappearance, and `test_excluded_crate_without_a_lockfile_is_skipped` currently codifies it. Since these seven files are checked in and consumed with `--locked`, derive the expected crate set first and report a missing lockfile (with an explicit exemption only if a genuinely lockless excluded crate is added).

tests/parity/cyclomatic_cross_language_parity.rs
```diff
@@ -105,23 +152,85 @@ fn assert_normalised(
...
+        let Some(row) = fixture(lang) else {
+            continue;
+        };
+        sums.insert(lang.name(), ccn_sum(lang, row.source, row.ext));
+        if row.offset != 0.0 {
+            offsets.insert(lang.name(), row.offset);
+        }
+    }
+
+    // A table that lost its rows is a real, feature-independent
```
dekobon, line 180: **Location:** `tests/parity/cyclomatic_cross_language_parity.rs:180`

**MEDIUM -- Do not let parity tests pass after measuring zero languages**

This deliberately permits `sums` to be empty, so every cyclomatic family reports success without parsing source or checking a metric. `cargo test -p big-code-analysis --no-default-features --test parity` currently lists and passes all 12 tests in this binary in 0.00s; the same `0 == 0` pattern occurs in the changed cognitive (`:198`, `:311`), exit (`:258`), and nargs (`:126`) drivers. That is the inverse of this PR's central feature-gating rule: a test with no eligible rows should be absent, not green. Please gate each test on the union of features represented by its `Some` rows and retain a non-vacuity assertion inside. The exhaustive `LANG` match still protects language-roster coverage; the `cfg` protects feature-subset execution coverage.

Makefile
```diff
@@ -1228,17 +1228,36 @@ fuzz-check:
...
+# target and still exits non-zero at the end.
+#
+# `fuzz-replay` below keeps the opposite shape deliberately: it is the
+# per-PR gate, where one broken seed is already enough to stop the
+# merge, and the seconds saved by not replaying the rest are the point.
 fuzz-smoke:
 	@if rustup toolchain list 2>/dev/null | grep -q '^nightly' && \
 	    command -v cargo-fuzz >/dev/null 2>&1; then \
 	  $(FUZZ_REQUIRE_SYMBOLIZER); \
+	  failed=""; \
```
dekobon, line 1247: **Location:** `Makefile:1247`

**MEDIUM -- Fail when target discovery fails or returns no targets**

The command substitution in `for t in $$(cargo +nightly fuzz list)` is evaluated in a loop word list, where this recipe's `-e` does not propagate a failing status. If `cargo fuzz list` errors or yields no targets, the loop runs zero times, `failed` remains empty, and `fuzz-smoke` exits 0 after fuzzing nothing. This gives the quarterly gate a false-success path precisely while adding its aggregate verdict. Capture discovery first (`targets="$$(cargo +nightly fuzz list)"`) so a nonzero status is fatal, reject an empty value explicitly, then iterate over `$$targets`.

Issue #1445, opened 2026-09-13
fix(metrics/loc): a bare backslash continuation row reads as blank in Bash


dekobon:
## Summary

A Bash line-continuation row that holds only a backslash reaches neither PLOC
nor CLOC, so `blank = sloc - ploc - cloc` claims it. The row is part of a
command — it is code — and nothing credits it, because `\` is spliced by the
lexer rather than surfaced as a leaf the `Loc` catch-all can see.

This is general to Bash, not specific to any construct.

## Reproduction

```console
$ printf 'echo a | \\\n\\\n  grep b\n' > /tmp/g1.sh
$ bash -n /tmp/g1.sh && echo VALID
VALID

$ bca metrics --no-config -O json -p /tmp/g1.sh | jq -c '.metrics.loc|{sloc,ploc,blank}'
{"sloc":3,"ploc":2,"blank":1}      # row 1 holds `\` and is code
```

Expected `ploc 3, blank 0`: all three rows are the one command.

A continuation row carrying *content* (`echo a | \` + `  grep b`) is fine — the
`grep b` leaves are credited by the catch-all. Only the bare-`\` row has no
leaf.

## Why it surfaced now

`src/metrics/loc/bash.rs`'s `heredoc_redirect` arm used to credit the wrapper's
whole interior, which incidentally covered continuation rows inside a heredoc's
command prefix. #1443 narrowed that range to the literal's own rows — correctly,
since the prefix is not part of the literal — and the mask went with it:

```console
$ printf 'cat <<EOT | \\\n\\\n  grep x\nbody\nEOT\n' > /tmp/k3.sh
$ bash -n /tmp/k3.sh && echo VALID
VALID
```

reports `ploc 4, blank 1` on `fix/batch-2026-09-12` against `ploc 5, blank 0`
before it. The defect is the same one as `g1.sh`; #1443 only stopped hiding it
in that one position, so this is not a regression to revert but a gap to close
in its own right. The bash arm carries a note pointing here.

## Root cause

`src/metrics/loc/bash.rs`'s catch-all is leaf-gated:

```rust
_ => {
    if node.child_count() == 0 {
        check_comment_ends_on_code_line(stats, start);
        stats.ploc.lines.insert(start);
    }
}
```

A bare-`\` row contains no node at all — the continuation is consumed by the
lexer joining the two physical rows into one logical one — so there is no leaf
to reach the arm and no container whose span the arm would credit.

## Fix sketch, unmeasured

[...]

Issue #1444, opened 2026-09-13
upstream(tree-sitter-bash): two heredocs on one command parse to an ERROR node


dekobon:
## Summary

`tree-sitter-bash` cannot parse two heredocs on one command. `cat <<A <<B` is
valid, runnable Bash; the grammar emits an `ERROR` node and misreads the second
`<<B` as `< <B` — a `file_redirect` — after which the second heredoc's body and
terminator are scored as ordinary commands.

Unlike #1443, whose shape is a bash syntax error and therefore unreachable in
runnable input, **this one is reachable**: the input runs.

## Reproduction

```console
$ printf 'cat <<A <<B\na\nA\n\nb\nB\n' > /tmp/two.sh
$ bash -n /tmp/two.sh && echo VALID
VALID

$ bca dump -p /tmp/two.sh
   ╰─ {heredoc_redirect:213} from (1, 5) to (3, 2)
      ├─ {<<:36}             from (1, 5) to (1, 7)
      ├─ {heredoc_start:152} from (1, 7) to (1, 8) : A
      ├─ {ERROR:65535}       from (1, 9) to (1, 10) : <
      ├─ {file_redirect:212} from (1, 10) to (1, 12) : <B
      ├─ {heredoc_body:218}  from (2, 1) to (3, 1)
      ╰─ {heredoc_end:156}   from (3, 1) to (3, 2) : A
   ├─ {command:207} from (5, 1) to (5, 2) : b
   ╰─ {command:207} from (6, 1) to (6, 2) : B
```

Rows 4 and 5 are heredoc B's body and terminator. They are scored as two
separate `command` nodes, so they reach `lloc` as logical lines and `ploc` as
code-by-command rather than as literal rows, and row 3 — blank, and inside B's
body in real Bash — is reported blank rather than as a literal interior row.

`bca metrics --no-config` reports `sloc 6, ploc 5, cloc 0, blank 1`. Under the
#1412 decision that a literal's interior rows are code, rows 3-5 all belong to
heredoc B, so `blank` should be 0 and `lloc` should not count `b` / `B`.

## Scope

This is an upstream grammar limitation, not a defect in our wrapper: there is no
arm to add, because the nodes the metric would need do not exist in the tree.
`AGENTS.md` asks that such cases be documented as upstream-grammar rather than
papered over, which is what this issue is for.

Options, none obviously right:

1. Report it to `tree-sitter-bash` and wait. Multiple heredocs on one command
[...]

Issue #1443, opened 2026-09-13
fix(metrics/loc): the Bash heredoc wrapper credits blank rows in a multi-row command prefix


dekobon:
## Summary

#1412 routes Bash's `heredoc_redirect` **wrapper** through
`add_string_interior_ploc`, which credits every row from the wrapper's
start to `end_line() - 1`. That is right for the body and the
terminator, which is what the fix was about. It is too wide on the other
side: the wrapper also spans the *command-line prefix*, and
tree-sitter-bash lets that prefix cross rows — `_heredoc_pipeline` is a
child of `heredoc_redirect`.

So a blank or comment-only row inside the prefix is credited as code.

## Reproduction

```console
$ printf 'cat <<EOT |\n\n  grep x\nbody\nEOT\n' > /tmp/hdp.sh

$ bca metrics --no-config -O json -p /tmp/hdp.sh | jq -c '.metrics.loc|{sloc,ploc,blank}'
{"sloc":5,"ploc":5,"blank":0}      # row 1 is blank
```

Before #1412 the same file reported `ploc 4, blank 1`. With a
comment-only row in place of the blank one, the row is reclassified from
comment-only to code-and-comment instead.

## Not executable, which is why this is not urgent

`bash -n` rejects that input: bash begins the heredoc body on the line
after the `<<` line, so the pipeline has no right-hand side. The valid
continuation spelling — `cat <<EOT | \` — keeps every wrapper row as
code, so it is unaffected and reports correctly today.

This is therefore grammar-reachable but not executable: the same
category as the C# `preproc_arg` shape #1430 deliberately leaves alone.
Filed so the gap is in CI's field of view rather than only in a review
transcript.

## Why no existing test catches it

`bash_heredoc_wrapper_credits_no_row_past_its_terminator`
(`src/metrics/loc.rs`) is #1412's upper-bound test, and it pins only the
rows *after* the terminator. Nothing bounds the prefix side, because
before #1412 the arm was keyed on the body and had no prefix to get
wrong.

## Fix sketch

Start the credited range at the `end_line()` of the last child before
`heredoc_body` / `heredoc_end`, rather than at the wrapper's
`start + 1`. Confirmed by hand to give the right answer on the four
shapes tried (bare `<<`, `<<-`, the pipeline above, and the valid
backslash continuation).

Weigh that against what it costs: it re-derives a grammar-shape
assumption that the comment block at `src/metrics/loc/bash.rs` spends
[...]

Issue #1431, opened 2026-09-13
fix(metrics/loc): exclude_tests leaves the #[cfg(test)] attribute row in ploc


dekobon:
## Summary

Under `--exclude-tests`, a file that is 100% test code still reports one
row of PLOC per pruned item: the `#[cfg(test)]` / `#[test]` attribute is
an `AttributeItem` **sibling** of the item it marks, not part of it, so
the prune never reaches it and Rust's `Loc` catch-all credits its start
row to PLOC.

Split out of #1417, which fixed the adjacent — and much worse — defect
where a shared row made `sloc` fall below `ploc`. This one produces a
correct-shaped but slightly-too-large reading, never a contract
violation, so it was deliberately kept out of that fix.

## Reproduction

Measured on `fix/issue-1417`, i.e. with #1417's fix already in place:

```console
$ printf '#[cfg(test)]\nmod t {\n    #[test] fn x() {}\n}\n' > /tmp/w.rs

$ bca metrics --no-config -O json -p /tmp/w.rs | jq -c '.metrics.loc|{sloc,ploc,blank}'
{"sloc":4,"ploc":4,"blank":0}

$ bca metrics --no-config --exclude-tests -O json -p /tmp/w.rs | jq -c '.metrics.loc|{sloc,ploc,blank}'
{"sloc":1,"ploc":1,"blank":0}      # a file with no production code at all
```

The attribute row is row 0; rows 1-3 are the `mod` item and are removed
correctly. Every pruned item costs one such row, so a `#[cfg(test)] mod
tests` at the foot of a production file inflates that file's `ploc` by
one, and a stacked attribute run (`#[cfg(test)]\n#[allow(…)]\nmod t`) by
one per row of the run.

## Root cause

`RustCode::should_skip_subtree`
(`big-code-analysis-ast/src/checker/rust.rs`) matches only
`ModItem | FunctionItem | ImplItem | TraitItem | ConstItem | StaticItem`.
`rust_item_is_test_only` *reads* the preceding attribute run to decide,
but the walker prunes the item alone — the attributes have already been
visited by the time it gets there, or are visited afterwards as ordinary
siblings, and either way each reaches
`src/metrics/loc/rust.rs`'s catch-all:

```rust
_ => {
    check_comment_ends_on_code_line(stats, start);
    stats.ploc.lines.insert(start);
}
```

Note this is PLOC only. SLOC is span-based and #1417 made its subtrahend
[...]

PR #1446, opened 2026-09-13
fix(metrics/loc): six LOC fixes from the 2026-09-10 residue

dekobon:
Six LOC issues, all residue of the 2026-09-10 batch (#1396 PHP heredoc, #1398
the line-set clamp) plus the older #1260. Three of the six were filed *by* that
work.

| Issue | Fix |
|---|---|
| #1414 | Helper-level tests pinning `add_multiline_string_ploc`'s parent gate (both directions) and `add_string_interior_ploc`'s `end_line()-1` bound. A *fourth* unguarded branch turned up and is covered too |
| #1412 | Bash `Loc` routes the `heredoc_redirect` wrapper, so a heredoc body that is empty at the top or throughout stops reading as blank |
| #1423 | The four C-family `PreprocArg` arms — and Python's open-coded copy, which the issue did not name — go through `add_string_interior_ploc`, bounding rows by `Node::end_line` |
| #1417 | `Sloc::excluded_lines` becomes a row *set*, settled against retained PLOC/CLOC at finalization, so `--exclude-tests` subtracts only rows the prune removes |
| #1431 | `--exclude-tests` prunes the `#[cfg(test)]` attribute with the item it marks |
| #1443 | A Bash heredoc no longer bills its command-line prefix as part of the literal |
| #1430 | Resolved as a **documented gap**: C# keeps no `PreprocArg` arm, with the reasoning recorded where its absence would otherwise read as an oversight |

## Notable

**`--exclude-tests` could report `sloc` below `ploc`** (#1417) — a contract
violation, not a rounding artifact. `Sloc` counted each pruned subtree's whole
row span, including rows retained code shares, on an assumption its own doc
comment spelled out and rustfmt-shaped input never violated. The fix carries the
pruned rows as a `LineSet` and subtracts only what the prune genuinely removed;
the `debug_assert!` at the clamp is tightened from the raw span to `sloc()`,
which makes the invariant a theorem rather than an observation and checks it on
every space of every walk.

**Three issue bodies were wrong in ways that changed the work.** #1412's
predicted snapshot churn does not exist (the submodule holds no shell files),
while #1423's — which its issue called unobservable — does, and moved
`left_test.cc`. #1417's expected numbers ignored that the `#[cfg(test)]`
attribute is a *sibling* of the pruned item.

**Two later commits are self-corrections.** `e10ed92b` restores the
`heredoc_body` arms that `cd439806` deleted: `node-types.json` describes the
well-formed grammar only, and under error recovery tree-sitter-bash emits an
orphan body with no wrapper, which reintroduced #1412's own symptom on three
valid inputs. `22e99725` corrects a reachability claim behind #1443 that was
measured too broadly, and files the valid-input gap that correction exposed
(#1445).

## Follow-ups filed

#1444 (upstream: `tree-sitter-bash` cannot parse two heredocs on one command),
#1445 (a bare `\` continuation row reads as blank — general to Bash, previously
masked inside heredocs).

## Verification

`make pre-commit` green on every commit. Patch coverage **97.61%** against
project **96.93%**, measured per-line over the added lines; the seven uncovered
are defensive branches and `debug_assert!` message arguments. `make
bench-scaling` all 28 probes within bound, `make chain-audit` clean. The
snapshot submodule is bumped to `98c48d2c` for the one DeepSpeech file #1423
moves.

Also adds lesson **#94**, merges a sub-example into **#84**, and adds four
entries to `.claude/rules/{testing,grammar-dispatch}.md`.


big-code-analysis-ast/src/checker.rs
```diff
@@ -728,6 +728,99 @@ fn rust_item_is_test_only<'a>(node: &Node<'a>, code: &[u8], ancestors: Ancestors
...
+/// the class is what #1100 was about.
+///
+/// Reusing [`rust_item_is_test_only`] rather than fusing the two walks
+/// into one pass is deliberate at that price. A fused scan would save
+/// one of the three sibling resolutions and would be a third reading of
+/// the same run, free to drift from the other two — the failure mode
+/// `.claude/rules/grammar-dispatch.md` §7 is about.
+fn rust_attributed_item<'a>(node: &Node<'a>, ancestors: Ancestors<'a, '_>) -> Option<Node<'a>> {
+    match ancestors.parent(node) {
+        Some(parent) if parent.child_count() <= forward_attribute_scan_budget(ancestors) => parent
```
dekobon: **Location:** `big-code-analysis-ast/src/checker.rs:807`

**HIGH -- Combined depth and attribute width makes this scan quadratic**

`should_skip_subtree` calls this once for every `AttributeItem`. In this branch each call starts again at the parent's first child, skips to the current attribute, and then scans to the decorated item. Because `forward_attribute_scan_budget` permits width proportional to depth (`3 * depth`), a parent at depth `D` with `3D` consecutive attributes takes this branch for every attribute and performs quadratic sibling work.

I reproduced this in release mode with valid Rust containing `D` nested functions followed by `3D` `#[cfg(test)]` attributes on one inner function. `bca metrics --no-config --jobs 1 --language rust --metrics nom --exclude-tests` took 0.04s, 0.18s, 1.00s, and 4.62s for D=250/500/1000/2000, while D=2000 without `--exclude-tests` took 0.02s. The existing depth-only and width-only probes cannot expose this diagonal shape. Avoid restarting from child zero for each attribute, and add a scaling probe where depth and one attribute run grow together.

src/metrics/loc/bash.rs
```diff
@@ -55,12 +102,67 @@ impl Loc for BashCode {
...
+            // to own: a continuation row holding only `\` has no leaf,
+            // so nothing credits it and it now reads blank. That is a
+            // general Bash gap, not a heredoc one — `echo a | \` + `\` +
+            // `  grep b` reports it outside any heredoc too — which the
+            // old blanket range happened to mask in this one position.
+            // Tracked as #1445; do not paper over it here.
+            HeredocRedirect => {
+                check_comment_ends_on_code_line(stats, start);
+                stats.ploc.lines.insert(start);
+                stats.ploc.lines.insert_range(
```
dekobon: **Location:** `src/metrics/loc/bash.rs:161`

**MEDIUM -- This turns valid continuation-only source into a blank row**

The new lower bound skips a valid Bash continuation row when that row contains only `\`. This input passes `bash -n`:

```bash
cat <<EOT | \
\
  grep x
body
EOT
```

At this head, `bca metrics --metrics sloc,ploc,blank` reports `sloc=5, ploc=4, blank=1`; the previous wrapper-wide range credited all five source rows. The comment above acknowledges the regression and defers the general case to #1445, but this PR newly makes an already-correct heredoc result wrong for valid input. Preserve continuation-only prefix rows here, or land the general continuation-row fix before narrowing this range, and add this valid fixture with `ploc=5, blank=0`.

src/metrics/loc.rs
```diff
@@ -11136,6 +11739,66 @@ class A {
...
+    ///   anchor available: deleting the dangling backslash also stops
+    ///   row 2 being credited, so a revert of the bound is the only
+    ///   evidence that half discriminates, and it was run.
+    #[test]
+    fn a_dangling_macro_continuation_does_not_credit_the_row_below() {
+        // Rows: 0-1 are the macro, 2 is the blank row its trailing
+        // backslash runs into, 3 is `int x;`.
+        const DANGLING_CONTINUATION: &[u8] = b"#define A 1 \\\n    2 \\\n\nint x;\n";
+
+        for lang in [
```
dekobon: **Location:** `src/metrics/loc.rs:11779`

**MEDIUM -- This regression test invokes disabled language parsers**

The table unconditionally includes C, C++, Mozcpp, and Objective-C. Under a feature subset that enables none of them, the test is still compiled and panics on its first row instead of being absent. Reproduced with:

```text
cargo test -p big-code-analysis --no-default-features --features go --lib a_dangling_macro_continuation_does_not_credit_the_row_below
... LanguageDisabled(C)
```

Gate the function on the union of the four features, gate each row on its own feature, and retain a non-vacuity assertion as required by the repository's feature-table test rule.

CHANGELOG.md
```diff
@@ -135,6 +135,96 @@ for historical reference.
...
+  the item never reached it and Rust's `Loc` catch-all credited its
+  start row to PLOC. A file that was nothing but test code reported
+  `sloc 1, ploc 1` — one row of code in a file with no production code
+  — and every `#[cfg(test)] mod tests` at the foot of a production file
+  inflated that file by one row, a stacked attribute run by one row per
+  attribute. `Checker::should_skip_subtree` now also answers for an
+  attribute, looking ahead over its run to the item it decorates and
+  pruning only when that item is one the prune removes anyway; the
+  attribute on a `use`, a `struct`, or any other kind `--exclude-tests`
+  keeps is untouched. **Metric drift:** `loc.ploc`, `loc.sloc` and the
```
dekobon: **Location:** `CHANGELOG.md:168`

**MEDIUM -- The public drift note omits Tokens and Halstead changes**

The new attribute arm `continue`s before `compute_per_node`, so it removes the attribute subtree from every selected metric, not only LOC. Rust attribute tokens are Halstead operators/operands, so `tokens` and the entire `halstead` block also fall; MI can move through Halstead volume as well as SLOC.

This is material drift rather than a theoretical detail: an independent base/head measurement over 463 Rust files found `tokens` and `halstead.length` changed in 206 files under `--exclude-tests`, with no change when the option is off. Please name `tokens` and `halstead.*` here and note both MI inputs, rather than limiting the documented drift to LOC rows.

dekobon:
## Code Review: fix(metrics/loc): six LOC fixes from the 2026-09-10 residue

**Verdict: CHANGES RECOMMENDED**

GitHub does not permit the PR author to submit a formal request-changes review on their own PR, so the findings are posted as comments.

| Severity | Count |
|----------|-------|
| BLOCKER | 0 |
| CRITICAL | 0 |
| HIGH | 1 |
| MEDIUM | 4 |

### Findings

| 1 | HIGH | `big-code-analysis-ast/src/checker.rs:807` | Combined depth and attribute width is quadratic | performance/security |
| 2 | MEDIUM | `src/metrics/loc/bash.rs:161` | Valid continuation-only prefix row becomes blank | correctness |
| 3 | MEDIUM | `src/metrics/loc.rs:11779` | C-family regression test ignores feature gates | test-gap |
| 4 | MEDIUM | `src/spaces_tests.rs:659` | Rust-only test module runs without Rust | test-gap |
| 5 | MEDIUM | `CHANGELOG.md:168` | Metric drift omits Tokens and Halstead | contract/docs |

### Review passes executed

- Correctness and logic bugs
- Language-dispatch and grammar-shape review
- Security and denial-of-service assessment (+ Semgrep; 0 findings)
- AST pruning and ancestor/sibling bookkeeping
- Performance and complexity scaling
- Test coverage and feature-matrix review
- Architecture, public contract, documentation, and snapshot consistency

Files reviewed: 23

src/spaces_tests.rs
```diff
@@ -1117,6 +1118,158 @@ impl Calc {
...
+    // `Loc` catch-all credited its start row to PLOC. A file that is
+    // nothing but test code read `sloc 1, ploc 1` — one row of code in a
+    // file with no production code — and every `#[cfg(test)] mod tests`
+    // at the foot of a production file cost that file one phantom row.
+    //
+    // `mi` is asserted to 0.0 on all three formulas because `sloc 0` is
+    // where `Mi::inputs_are_empty` takes over, and the issue raised that
+    // as a hazard: it is not one — a file with nothing in it scores
+    // zero, deliberately and for every formula.
+    #[test]
```
dekobon: **Location:** `src/spaces_tests.rs:659`

**MEDIUM -- The Rust-only module runs when the Rust feature is disabled**

This module constructs `RustParser` unconditionally. The new tests therefore fail under supported feature subsets without Rust; for example:

```text
cargo test -p big-code-analysis --no-default-features --features go --lib an_all_test_file_measures_zero_rows
... LanguageDisabled(Rust)
```

Add `#[cfg(feature = "rust")]` to the module so these tests are absent rather than spuriously failing when their parser is not compiled. The new Rust-specific tests around `rust_loc_pruned` in `src/metrics/loc.rs` need the same gate.

big-code-analysis-ast/src/checker.rs
dekobon: Confirmed and fixed in `9c21c7e3`. Reproduced worse than reported here — 0.08 / 0.34 / 1.21 / **5.18 s** for D=250/500/1000/2000 against 0.05 s with the flag off.

Both causes turned out to be one: every attribute in a run decorates the same item, so all L answers are equal, and each ask re-derived the run twice (`skip_while` from child 0, then again inside `rust_item_is_test_only`). Rather than making the scan cheaper, a verdict now reports how far it reaches — `should_skip_subtree` returns `SubtreeSkip` instead of `bool`, the attribute arm sets the reach to the decorated item's start byte, and the walk reuses the last reaching verdict. Pre-order visits `start_byte` non-decreasing, so the test is O(1) and a passed reach cannot be re-entered. One ask per run.

Same input now: **0.00 / 0.01 / 0.02 / 0.07 s**.

`forward_attribute_scan_budget` is unchanged — measured, the forward pass is still the cheaper of the two readings on the diagonal, and the budget was never the cause. `rust_item_is_test_only` is reused verbatim, so no third reading of the run was introduced (§7).

New probe `nom/deep-attribute-run` grows depth and one run together. Watched failing: **2.01, over bound** with the reach disabled; **1.06** with it. Its reading is `nom.total()` of the retained wrappers rather than anything the prune produces, so a later change to what is pruned cannot zero it.

One you did not raise and I am not fixing here: the *other* diagonal — D nested fns with 3D **separate** one-attribute items, runs of length 1 so nothing to reuse — is still quadratic (3.07 s at D=1000). About half predates #1431; it is the item path's own `rust_outer_attr_marks_test` restarting per item. Filed separately.

src/metrics/loc/bash.rs
dekobon: You are right, and it changes the disposition: **the narrowing is reverted** in `9c21c7e3`.

Neither remedy you offered was available. Telling a `\`-only row from an empty one needs the source bytes, and `Loc::compute` is not given them, so "preserve continuation-only rows here" cannot be written; and landing the general fix first means widening the `Loc` trait across twenty languages.

So this falls to the rule the sibling #1430 note in this same PR states: fix only when it costs nothing on valid input. It costs something here — what the narrowing fixed needs `bash -n` to reject the file, what it broke runs — so #1443 becomes a documented gap like #1430 rather than a fix. The arm now records why the narrowing looks right and is still wrong, so a second attempt fails review rather than shipping.

The fixture is inverted to your suggested values: `bash_heredoc_continuation_row_is_code` pins `ploc 5, blank 0` on exactly your input, plus the content-carrying continuation as the bracket. General gap tracked as #1445.

src/metrics/loc.rs
dekobon: Fixed in `9c21c7e3`. Gated on `#[cfg(any(feature = "c", feature = "cpp", feature = "mozcpp", feature = "objc"))]`, the hardcoded array replaced by a per-row `#[cfg]` table, and `assert_fixtures_present` kept for non-vacuity — both halves, per the rule.

Verified with your reproducer: under `--features go` the test is now **absent** (`0 passed; 0 failed`), not failing and not vacuously passing; under `--features cpp` it runs on the single enabled row.

Also fixed the sibling `a_continued_macro_body_counts_every_row_it_spans`, which has the identical shape two lines away. That one is pre-existing rather than introduced here, but it is the same edit and leaving it would have made the file inconsistent with itself.

CHANGELOG.md
dekobon: Fixed in `9c21c7e3`, and you are right that it is material rather than theoretical — I had scoped the note to LOC because that is what the issue was about, not because I had checked.

Measured on one fixture to confirm the mechanism before rewriting: `tokens` 33→6, `halstead.length` 22→4, `halstead.volume` 83.8→8.0, `mi.original` 115.8→148.5.

The entry now says the prune arm `continue`s before the per-node computes so the attribute subtree leaves *every* selected metric, names `tokens` and the whole `halstead` block, states that MI moves through **both** of its inputs rather than SLOC alone, and carries your 206-of-463 figure.

src/spaces_tests.rs
dekobon: Fixed in `9c21c7e3`. `#[cfg(feature = "rust")]` on the `exclude_tests_rust` module, and the same gate on `rust_loc_pruned` and its three callers in `src/metrics/loc.rs` — helper and callers together, so no unused-function warning appears under a subset.

Verified: all six named tests absent under `--features go`, with zero warnings in that build.

One deliberate omission — the older `rust_loc` helper and its 28 callers are also ungated. That is pre-existing (#1051 era) and part of the #1285 pile, so I left it rather than widening this PR; say the word if you would rather it came along.

dekobon:
## Review findings addressed — `9c21c7e3`

All five, pushed. Gate green (`BCA_GATE: pass`), and each fix verified against the reviewer's own reproducer rather than by inspection.

| # | Severity | Disposition |
|---|---|---|
| 1 | HIGH | Fixed — 5.18 s → 0.07 s at D=2000, new diagonal probe watched failing |
| 2 | MEDIUM | **Reverted** the #1443 narrowing rather than patching it |
| 3 | MEDIUM | Gated, absent under `--features go`; pre-existing sibling fixed too |
| 4 | MEDIUM | Gated, six tests absent, zero warnings |
| 5 | MEDIUM | Drift note now names `tokens`, `halstead.*` and both MI inputs |

Two are worth a second look because they changed the shape of the PR rather than just patching it.

**Finding 1** turned out to have one cause, not two: every attribute in a run decorates the same item, so all L answers are equal and the run was being re-derived L times. The fix makes a verdict report how far it reaches — `should_skip_subtree` returns `SubtreeSkip` instead of `bool` — so the walk asks once per run. That is a cross-crate signature change on a trait explicitly outside the stability contract, with one override, one call site and one direct test; flagging it rather than burying it. `forward_attribute_scan_budget` is untouched, having been measured not to be the cause.

**Finding 2** reverses #1443 outright. Neither remedy offered was writable — distinguishing a `\`-only row from an empty one needs source bytes `Loc::compute` does not receive — and by the rule the #1430 note states in this same PR (fix only when it costs nothing on valid input), the narrowing should never have landed. It is now a documented gap like #1430.

Filed from this round: **#1447**, the adjacent diagonal that remains quadratic — D nested fns with 3D *separate* one-attribute items, where runs of length 1 leave nothing to reuse. About half predates #1431; it is #1100's budget design on the item path. Neither the fix nor the new probe covers it, and the issue says so.

Also carried over: **#1445** (a bare `\` continuation row reads as blank, general to Bash) and **#1444** (upstream: two heredocs on one command).

Issue #1458, opened 2026-09-14
meta: sibling sweeps are asserted rather than measured


dekobon:
## Summary

A meta-observation from the 2026-09-13 batch (#1420, #1406, #1422, #1421,
#1407, #1410), worth recording because the pattern is about how these issues
get *found and filed*, not about any one fix.

**Five of the six fixes were correct in the language they targeted and
unswept to at least one sibling that shares the defect.** Each issue carried
a sibling-sweep section; each was either absent, incomplete, or — twice —
actively wrong.

## The seven gaps the batch left behind

| gap | issue | how it surfaced |
|---|---|---|
| Ruby subject-less `case` double-counts | #1453 | #1421's body said Ruby was swept clean; two agents independently measured otherwise |
| Java 21 / Rust / Python match guards | #1454 | #1422's own "Rust is the convention" argument is false for Java |
| Elixir / Ruby guards | #1454 | whole-branch review |
| Objective-C `@available` | #1457 | #1410 surveyed this neighbourhood and excluded the wrong node |
| Comment in a condition slot zeroes it | #1455 | #1422 wrote its own slot extras-safe and left the siblings |
| `goto case` counts as a decision | #1451 | #1420 had to enumerate what emits `Case` |
| `is_call` vs `abc.branches` | #1456 | #1406 |

Plus three found *within* the batch and fixed there, all of the same class:
C#'s `catch_filter_clause` (#1422's body named only `when_clause`), C#'s bare
`is` type test (the Kotlin half landed in the same branch and the C# half did
not), and Kotlin's infix `and` / `or` / `xor`.

## What is actually going wrong

`.claude/rules/grammar-dispatch.md` already closes with "when you fix one
language, sweep the rest", and every issue body has a sibling-sweep heading.
The rule is being followed as *prose* and not as a procedure:

1. **Sweeps are asserted rather than measured.** #1421's body listed four
   languages as clean "measured as part of #1383 … so no need to re-derive".
   Ruby was not clean. Nothing distinguishes a measured row from a remembered
   one once it is written down.
2. **The sweep is scoped to the construct the issue names, not to the
   predicate being changed.** #1410 swept for numeric literals and missed
   `available_expression`, which is not numeric but occupies the same slot in
   the same set.
3. **Two fixes in one batch can share a defect and not see each other.** The
   C# and Kotlin halves of the bare-type-test gap landed in the same branch,
   four commits apart, and only a whole-branch read connected them.

## Suggested changes

[...]

Issue #1457, opened 2026-09-14
fix(abc/objc): `@available` scores no condition


dekobon:
## Summary

Objective-C's `@available` check scores no ABC condition. It is the same
within-one-language asymmetry [#1410](https://github.com/dekobon/big-code-analysis/issues/1410)
closed for numeric literals, one kind over.

## Measured

```
if (@available(iOS 13.0, *)) { … }   →  abc.conditions 0,  cyclomatic 2
if (a) { … }                         →  abc.conditions 1,  cyclomatic 2
```

`bca dump` shows `if_statement > parenthesized_expression >
available_expression`, so the condition slot **is** reached — the kind is
simply unlisted.

## Where

`cpp_bool_terminal_kinds!` (`big-code-analysis-ast/src/macros/kind_sets.rs`).
The set is name-keyed and already carries `identifier`, `call_expression`,
`message_expression` and `field_expression`; `available_expression` (486) is
missing.

Note the set is shared by C, C++, Mozcpp and Objective-C, so the arm would be
inert for the first three — `@available` is an Objective-C construct. That is
the same situation `message_expression` is already in.

## Why it is filed now rather than found earlier

#1410 surveyed exactly this neighbourhood and wrote a conclusion about it:
its comment records that Objective-C's `version_number` is deliberately
excluded as "the `@available(iOS 13.0, *)` token, not a literal, and never
occupies an operand slot". That is true of `version_number` and rules out the
wrong node — the *wrapper* `available_expression` is the operand. So the gap
now reads as audited-and-accepted rather than unnoticed, which is worse than
it being missed.

## Tests

An `@available` check alongside a plain identifier condition in the same
fixture, so the gate is anchored on a second axis; and a C or C++ control
confirming the shared set is unaffected.

## Resolution

Fixed on `fix/issue-1457` (commit `bf6252bc`); closes on merge.

`"available_expression"` joins `cpp_bool_terminal_kinds!()`
(`big-code-analysis-ast/src/macros/kind_sets.rs`), the name-keyed
terminal set that the C-family ABC walkers share. Objective-C owns no
set of its own — `src/metrics/abc/objc.rs` imports the C++ helpers — so
the one entry is the whole fix.

The defect was wider than the reported row. Measured before/after with a
release build, `abc.conditions` rises from 0 to 1 for the `if`
predicate, the `__builtin_available` spelling, a `for` header condition
and a negated `while` condition, and from 1 to 2 for a `&&` operand and
[...]

Issue #1456, opened 2026-09-14
chore(csharp): reconcile is_call with abc.branches, and primary-constructor space attribution


dekobon:
## Summary

Two small divergences noticed while fixing
[#1406](https://github.com/dekobon/big-code-analysis/issues/1406), neither
introduced by it, both worth a decision rather than being left implicit.

### 1. `Checker::is_call` disagrees with `abc.branches`

C#'s `Checker::is_call` matches only the invocation aliases. `abc.branches`
counts those plus `ObjectCreationExpression`, `ConstructorInitializer`, and
since #1406 the primary-constructor base call
(`PrimaryConstructorBaseType` / a `base_list`-gated `ArgumentList`).

So `bca find --filter call` and `abc.branches` now disagree by one more
construct than they used to. `.claude/rules/grammar-dispatch.md` §7 asks for
exactly this cross-walk: two predicates classifying the same nodes through
parallel `matches!()` drift silently, and each reads as internally
consistent from its own side.

Decide which is intended — `is_call` is arguably answering a narrower
question ("is this a call site a reader navigates to") than ABC's
Fitzpatrick branch rule ("function invocation or object construction"). If
the divergence is deliberate, say so in a comment at both sites; if not,
widen `is_call`.

### 2. A primary constructor opens no `FuncSpace`

`class Sub(int x) : Base(x)` has no constructor body, so no function space
is opened and the base call lands on the **class** space. The classic
`: base(x)` spelling lands on a function space.

File-level sums agree, so nothing is wrong at the aggregate. But
per-function `branches_max` / `branches_average`, and a per-function
`bca check --threshold abc=N`, see the two spellings differently — the
primary-constructor form contributes to the class row and never to any
function row.

That may well be correct (there is no function there to attribute it to),
but it means a codebase that migrates to primary constructors will see its
per-function ABC distribution shift without any behaviour change. Worth
recording the decision either way.

## Related

- #1406 — where both were noticed.
- #1279, #1384 — the sibling delegation fixes.

dekobon:
## Fix Summary

Both divergences are real and both were verified before deciding. Neither
is a bug, and neither metric nor predicate changed: this lands the two
decisions as documentation, cross-references at both sites, and tests
that pin them.

### 1. `is_call` narrower than `abc.branches` — deliberate, and not C#-specific

This is the workspace convention, not a C# quirk. Surveying every
`Checker::is_call` against its language's ABC branch arms:

| Relation | Languages |
[...]

Issue #1455, opened 2026-09-14
fix(abc): a comment in a condition slot zeroes the condition


dekobon:
## Summary

C#'s `if` / `while` / `do` condition slots read a **fixed child index**, so
a leading comment inside the parentheses shifts the read and the condition
scores zero.

```console
$ # if (IsEven(x))      -> abc.conditions 1
$ # if (/*c*/ IsEven(x)) -> abc.conditions 0
```

This is the same class [#1181](https://github.com/dekobon/big-code-analysis/issues/1181)
fixed for the ternary, and the same trap the `when`-guard slot added in
[#1422](https://github.com/dekobon/big-code-analysis/issues/1422)
deliberately avoids: tree-sitter `extra`s are **named** nodes, so a
first-named-child read hands the slot a `comment`. That arm iterates every
named child instead, and `csharp_guard_keeps_its_condition_across_a_comment`
pins it. The sibling slots were left to their own change rather than
widened into #1422.

## Where

`src/metrics/abc/csharp.rs`, `csharp_walk_for_conditions` — the
`IfStatement | WhileStatement` and `DoStatement` arms, which use
`csharp_inspect_child(node, idx, …)` with a positional index. The guard arm
immediately below them shows the shape that survives extras.

## Fix shape

Prefer `child_by_field_name` where the grammar exposes a field for the slot
(`.claude/rules/grammar-dispatch.md` §3). Where it does not, iterate named
children and pick by role, as the `WhenClause | CatchFilterClause` arm does.

Check the sibling languages in the same pass — `java.rs`, `groovy.rs` and
`cpp.rs` all carry positional `inspect_child` reads, and Groovy's
`if` / `while` (`child(2)`) and `do`-`while` (`child(4)`) are called out in
`src/metrics/abc/groovy.rs` as the last positional condition lookups after
#1181/#1276 migrated the rest.

## Tests

A commented and an uncommented spelling of the same condition, asserted
**against each other** rather than against a literal, so the pair cannot
drift apart silently. Both block (`/* */`) and line (`//`) comments, since
they are separate extras.

## Related

- #1181 — the ternary instance.
- #1422 — where the guard slot was written extras-safe and the sibling
  slots deliberately were not.

PR #1459, opened 2026-09-14
fix(metrics): six ABC and cyclomatic fixes

dekobon:
Fixes six ABC / cyclomatic issues, plus two defects found reviewing the
batch as one diff.

| Issue | Commit | What it does |
|---|---|---|
| #1420 | `f04a0075` | C# comparison-operator overloads stop scoring a spurious condition |
| #1406 | `268b9760` | C# primary-constructor base call counts as a branch |
| #1422 | `eb4468cb` | A C# `when` guard is a decision in cyclomatic and a condition slot in ABC |
| #1421 | `605c771e` | Kotlin subject-less `when` arms stop double-counting their comparison |
| #1407 | `ef20b2ca` | Enum constants with arguments count as a branch in Kotlin / Java / Groovy |
| #1410 | `39e01116` | Numeric operands count in PHP, Groovy and the C family |

All six move published metric values; each CHANGELOG entry carries a
**Metric drift** line.

## Verification

Each commit passed a full `make pre-commit` independently, with the verdict
read from the `BCA_GATE:` line. Patch coverage is **100.00%** (740/740 added
instrumented lines) against a project 95.58% regions / 96.96% lines. Every new
match arm was verified by perturbation — exact inverse edits, never a
file-level restore — with the harness asserted non-vacuous first.

Two submodule bumps are pushed and recorded in the same parent commit as their
fix: `cf1ef381` (one C# snapshot, #1422) and `ee1bd859` (seven DeepSpeech
snapshots, #1410). Across all seven of the latter the only sections that move
are `abc`; across the former, only `cyclomatic` and its derived metrics.

## Two fixes the whole-branch review added

Per-issue review and six green gates missed both; only reading the commits
together exposed them.

- **`d08eb8e9`** — `csharp_bool_terminal_kinds!()` listed `IsPatternExpression`
  (392) but not `IsExpression` (391). They are distinct kinds: the grammar
  emits the second only once a pattern is involved. So `if (x is int)` scored
  0 conditions against a cyclomatic decision of 1 while `if (x is int y)`
  scored 1 — and #1422's guard was **still spelling-dependent** in exactly the
  case its own comment claimed to have fixed. This is the C# half of the gap
  #1421 closed for Kotlin four commits earlier; neither fix swept toward the
  other.
- **`ac59a1a3`** — a regression #1421 introduced. Kotlin spells boolean `and` /
  `or` / `xor` as infix *functions*, so `a and b` is an `infix_expression`
  matched by no token arm, and the blanket per-entry count that fix removed had
  been covering it. `when { a and b -> … }` fell from 1 condition to 0 while
  the `if (a and b)` it is supposed to agree with still scored 1 — that
  contradiction is what made it diagnosable. Third shape the blanket count
  turned out to be propping up, after `is` / `in` and bare parens.

## Issue bodies that were wrong

Worth flagging for anyone reading the issues alongside this branch — each was
re-verified against a live build before any code was written:

- **#1421** proposed a fix that cannot be written (`WhenCondition` maps to
  `"_when_condition"`, a hidden rule the parser never emits), and listed Ruby
  as swept clean when Ruby has the defect verbatim (#1453).
- **#1410** gave a corpus rationale for Groovy that is false — no corpus carries
  a `.groovy` file — and its alias warning pointed at a trap that is not present
  while missing the one that is (PHP's `Float2` is a *type keyword* rendering to
  the same string). Both stale comments are corrected in the diff rather than
  deleted.
- **#1422** described only half the defect; `catch (E e) when (…)` is a separate
  grammar node with the same problem, fixed here too.
- **#1407** left its central question open; it is decided and the rationale is
  recorded on each arm.

## Two things to look at before merging

1. **#1421 required a scope expansion.** Fixing it meant adding `is` / `in` /
   infix calls to Kotlin's terminal set, so `if (a is String)` goes 0 → 1
   condition. Without it the fix regresses four shapes. Documented in the
   commit and CHANGELOG.
2. **`.bca-baseline.toml` carries one unrelated line** — `metrics_inner`'s
   halstead effort, stale on `main`, a decrease that rode along with a
   wholesale regeneration. Named in `ef20b2ca`'s message.

## Follow-ups filed

#1451, #1453, #1454, #1455, #1456, #1457 — sibling gaps this batch did not
close, each with measurements.

#1458 is the meta-issue: five of six fixes were correct in their target
language and unswept to a sibling with the same defect, and every issue body
had a sibling-sweep section that was absent, incomplete, or twice actively
wrong.


dekobon (commented):
## Code Review: fix(metrics): six ABC and cyclomatic fixes

**Verdict: REQUEST CHANGES**

| Severity | Count |
|----------|-------|
| BLOCKER | 0 |
| CRITICAL | 0 |
| HIGH | 2 |
| MEDIUM | 0 |

### Findings

| 1 | HIGH | `src/metrics/abc/kotlin.rs:238` | Preserve the subject-less `when` arm decision for non-terminal predicates | correctness |
| 2 | HIGH | `src/metrics/abc/csharp.rs:534` | Count the guard slot when its expression already has operator structure | correctness |

### Review passes executed

- Correctness and logic bugs
- Grammar-dispatch and cross-language parity
- Regression-test discrimination
- Performance and scalability
- Metric/documentation consistency
- Semgrep static analysis (0 findings)

Files reviewed: 12

All GitHub CI checks are green. This review is submitted as a comment because GitHub does not allow an author to request changes on their own pull request.

src/metrics/abc/kotlin.rs
```diff
@@ -152,6 +165,81 @@ fn kotlin_count_condition(condition: &Node, parent: &Node, conditions: &mut f64)
...
+// `when { x, y -> … }` off that count rather than onto it. Alternatives
+// after the first are still seen by the token arms.
+fn kotlin_count_when_entry<'a>(
+    entry: &Node<'a>,
+    ancestors: Ancestors<'a, '_>,
+    conditions: &mut f64,
+) {
+    if kotlin_enclosing_when_has_subject(entry, ancestors) {
+        *conditions += 1.;
+    } else if let Some(condition) = entry.child_by_field_name("condition") {
```
dekobon, line 307: **Location:** `src/metrics/abc/kotlin.rs:238`

**HIGH -- Preserve the subject-less `when` arm decision for non-terminal predicates**

Delegating the arm entirely to `kotlin_count_condition` drops the decision whenever the condition is a valid boolean expression outside that helper's terminal/paren/unary allowlist. For example, on this head `when { a!! -> 1; else -> 0 }` and `when { a as Boolean -> 1; else -> 0 }` both report `abc.conditions = 0` with `cyclomatic = 2`; at the base revision both report one ABC condition. `when { a ?: false -> 1; else -> 0 }` similarly falls from 2 to 1 while cyclomatic remains 3, because only the Elvis operator survives. The removed blanket `WhenEntry` count supplied the arm decision in all three cases. Please retain one decision for subject-less entries whose expression walker does not already account for the slot, and add regressions for at least one such wrapper/operator shape.

src/metrics/abc/csharp.rs
```diff
@@ -404,6 +490,52 @@ fn csharp_walk_for_conditions<'a>(
...
+        // `!`-prefix wrapper — so passing them through the slot adds
+        // nothing and the loop cannot double count a clause that holds
+        // one expression by construction.
+        //
+        // FIXME(#1455): the sibling `if` / `while` / `do` slots read a
+        // fixed child index and so still lose their condition to a
+        // leading comment (`if (/*c*/ g)` scores 0). That is the same
+        // class of bug and predates this arm; it is left to its own
+        // change rather than widened into here.
+        WhenClause | CatchFilterClause => {
```
dekobon, line 562: **Location:** `src/metrics/abc/csharp.rs:534`

**HIGH -- Count the guard slot when its expression already has operator structure**

This traversal only adds the new guard-slot condition when `csharp_count_condition` recognizes the guard node as a terminal or wrapper. A valid guard such as `true when b ?? false => 1` has a `BinaryExpression` child, so the new arm contributes nothing; the function reports `abc.conditions = 1`, entirely from the non-default switch arm, while cyclomatic is 4 (base cyclomatic was 3 and ABC was already 1). Thus #1422 raises cyclomatic for the guard but leaves ABC unchanged, despite this branch's stated goal that every guard spelling contributes one slot. The pre-existing omission of `QMARKQMARK` from ABC does not account for the missing guard decision: even if `??` remains out of scope, this clause still needs its own +1. Please add the slot independently of any nested operator counts and cover a guard whose root is a non-terminal binary expression.

src/metrics/abc/kotlin.rs
dekobon: Fixed in `b2b78598`. Reproduced first — `when { a!! -> … }` and `when { a as Boolean -> … }` both scored 0 against a decision count of 1, exactly as reported.

**Root cause** was narrower than "the slot drops non-terminal predicates": `unary_expression` is **one kind for both Kotlin unary spellings**, so the slot routed `a!!` in as *handled*, and then `kotlin_inspect_container`'s positional `child(1)` read — correct for the prefix `!x` — found the `!!` token instead of the operand and stopped. `as_expression` was not recognised as a wrapper at all.

**Fix**: extracted `kotlin_wrapper_operand`, which returns the operand a transparent wrapper wraps plus whether the wrapper proves booleanness. It handles parens, prefix `!`, postfix `!!` and a plain `as` cast, and the peel loop chains them, so `(a!! as Boolean)` peels three wrappers to one `identifier`.

Two changes beyond the minimum, both aimed at the class rather than the instance:

- Operands are now read by grammar **field** (`operator`/`argument`, `left`/`right`), so one read serves every spelling of a kind. A positional index is correct for exactly one spelling and silently wrong for the rest — which is what produced this defect.
- `kotlin_count_condition` now **derives** its wrapper set from the peel instead of restating it. Those two lists disagreeing *is* this bug, so there is now one source of truth (§7 applied within one language).

**`as?` deliberately excluded.** `AsQMARK` is already a condition token in the same file. Measured: peeling `as?` too scores `a as? Boolean` **2** against every other spelling's 1 — a §5 double count. So `as` is owned by the peel and `as?` by its token, and both read 1. Pinned by a `(a as? Boolean)!!` member, which reaches the cast through two other wrappers using valid Kotlin.

**Measured after** (control `a` = 1/2): `a!!` 0→1, `(a!!)` 0→1, `a as Boolean` 0→1, `a!! as Boolean` 0→1, `a as? Boolean` 1→1 (unchanged), `a && b` 2→2. Controls `a > 5` and `g(a)` unmoved at 1.

**Regressions requested**: `kotlin_condition_slot_peels_null_assertions_and_casts` (eleven members, per-member conditions+cyclomatic table, kind-count anchors) and `kotlin_chain_operands_peel_null_assertions_and_casts` — `kotlin_inspect_container` has two callers and every other fixture reached it through the slot, so the `&&`/`||` walker needed its own coverage. Both revert-verified; the second was anchored in `839b6727` after review found it could decay (plain `a && b` scores the same 2/3, so trimming the `!!` would have left it green with its subject gone).

No Kotlin corpus, so no snapshot movement.

src/metrics/abc/csharp.rs
dekobon: Fixed in `b2b78598`, though **not** by the mechanism suggested — the measurement argued against it, so I want to lay that out rather than quietly do something else.

**Reproduced.** Holding the fixture shape fixed and varying only the guard (`class A { static bool E(int x)=>true; int f(int x, bool? b) => x switch { > 0 when <G> => 1, > 0 => 2, _ => 3 }; }`):

| guard | conditions | cyclomatic |
|---|---|---|
| `x % 2 == 0` | 3 | 7 |
| `x > 2` | 3 | 7 |
| `E(x)` | 3 | 7 |
| `x is int` | 3 | 7 |
| `b ?? false` | **2** | **8** |

So the `??` guard contributed 0 where every other spelling contributed 1 — and its cyclomatic is *higher*, so the two metrics diverged in opposite directions.

**Why not an unconditional slot increment.** Measured both readings of "add the slot independently of any nested operator counts":

- Slot always `+1`: `when x > 2` goes 3 → **4**, reintroducing precisely the double count #1422 removed.
- Slot `+1` with suppression inside the guard: `when a > 1 && b < 2` goes 4 → **3**, flattening the compound sub-structure #1422 deliberately preserves.

Uniformity and sub-structure conflict under that reading. They stop conflicting once the real gap is closed.

**The actual gap** is that `??` is a C# **cyclomatic** decision (`cyclomatic/csharp.rs:86`) counted by **no ABC arm** — a pre-existing omission the code already acknowledged in a comment at `abc/csharp.rs:403`. The guard slot correctly declined a `binary_expression`, and then nobody counted its operator. Adding `QMARKQMARK` to the condition-token arm levels every spelling at 3, keeps the compound guard at 4, and touches no slot.

You are right that the `??` omission "does not account for the missing guard decision" as a *justification* — but it does turn out to be the whole mechanism. And it fixes `??` everywhere rather than only inside a guard: `int y = x ?? 0;` goes 0 → 1, now at exact parity with cyclomatic.

Two supporting facts verified rather than assumed: a `grammar.json` sweep of the pinned tree-sitter-c-sharp 0.23.5 finds `??` in `binary_expression` and nowhere else, so the arm needs no gate; and the JS/TS family has always counted `QMARKQMARK` (`js_family.rs:356`), so this closes a cross-language divergence rather than opening one.

**Honest limit, now stated in the code**: this narrows the ABC-versus-cyclomatic gap on a `??` guard from two to one without closing it. Cyclomatic counts `??` *in addition to* the guard clause, while ABC's condition substitutes for the slot — so `when b ?? false` reads 3 against cyclomatic 5. The residual is the slot's standing policy of leaving a `binary_expression` to its operators; `when x > 2` pays the same and merely breaks even. Away from a guard, parity is exact. `??=` is unchanged and remains an assignment, matching the JS family.

**Coverage requested**: `csharp_switch_arm_guard_scores_one_condition_however_spelled` gained a `nullc` member rather than a parallel test, so the guard-spelling table is now six members and a future change that fixes one spelling while breaking another fails here. Added `csharp_null_coalescing_is_a_condition` for `??` away from any guard, where nothing else can supply the count.

Costs one corpus snapshot — `linq.cs` is the only C# corpus file with a `??`; file-level conditions 10 → 12, abc keys only. Submodule `81abf593`.

Issue #1478, opened 2026-09-16
test(features): nothing detects an over-narrow per-language test gate


dekobon:
## Summary

Nothing detects a per-language test gate that is too **narrow**.

#1472 gated ~3,235 test items on the grammars they name. Two safety nets
check that work, and both only catch the *wide* direction:

- `cargo nextest list --all-features`, diffed before and after, proves no
  test left the all-features build. An over-narrow gate still runs there,
  so it is invisible.
- The `feature-matrix` legs now run the suite. A gate that is too wide
  panics; a gate that is too narrow simply drops the test, and a leg that
  runs 200 fewer tests looks exactly like a green run.

So a future `--fix` pass — or a hand-edited marker — can silently stop a
test running on every partial leg with nothing to say so. That is the
failure mode #1472's own commit messages name as the dangerous one.

## Resolution

Both candidates below were built, because they answer different
questions and neither subsumes the other.

**`over_gated`, in `check-test-lang-gates.py`.** For each item, the
conjunction of its own `cfg` is measured against the languages its body
reaches. A feature is over-declared when turning *it alone* off is what
excludes the item, and the item needs neither it nor anything that
enables it — measured from a build that satisfies whatever the gate
wants off, so a `cfg(not(feature = "javascript"))` disabled-path test is
read from a build it actually compiles in.

The opt-out is a marker rather than a baseline, because the population
turned out small enough for one: **16 items**, in four shapes, each
already carrying prose explaining itself.

```rust
// test-lang-gates: hand-written(cpp) — the corpus walk picks a
//     language per file from its extension, so the glob list decides
//     it and nothing in the body names it
#[cfg(feature = "cpp")]
```

Naming the features is load-bearing: a marker for the wrong one does not
silence the real one. A marker that stops being load-bearing fails too —
otherwise the sixteen stop reading as a census and start accumulating.

**`--compare <ref>`**, for the residue. `over_gated` compares a marker
against the derivation; when the two agree and are *both* wrong, only
the previous revision says so. That turned out to be computable
statically: scan the tree at `ref` as well (`git archive` into a scratch
directory — no cargo, no second build), derive which single-language
builds compile each test in each, and fail on a test that still exists
but stopped being built somewhere. The probe set comes from the language
table, so a new language extends it for free.

The `lint` CI job runs it per pull request against the merge base.
Deliberately not in `make pre-commit`, which has no base revision to be
meaningful against. A *deliberate* narrowing looks the same from here
[...]

Issue #1285, opened 2026-08-08
test(features): no partial-feature build can run the test suite


dekobon:
Found by a feature-matrix audit of the root library crate at d86d3806. Every leg below was built and run sequentially in a clean worktree; each command and its full log path is quoted so the numbers can be reproduced.

## Finding

No partial-feature build of `big-code-analysis` can run the test suite. Every leg fails with thousands of test panics, almost all of the same shape — an ungated per-language test constructs a parser for a language whose feature is off and hits the `Tree::new` invariant:

```
thread '...' panicked at src/node.rs:71:52:
invariant: dispatcher cfg-gates this call behind the per-language Cargo feature: LanguageDisabled(Python)
```

Measured (`cargo nextest run --no-default-features [--features …] -p big-code-analysis`):

| feature set | result |
|---|---|
| *(none)* | 3126 run: **2598 failed**, 528 passed, 8 skipped |
| `rust` | 3176 run: **2380 failed**, 796 passed |
| `go` | 3128 run: **2498 failed**, 630 passed |
| `rust,typescript` (the CI minimal-langs subset) | 3183 run: **2124 failed**, 1059 passed |
| `tcl,elixir` | 3131 run: **2406 failed**, 725 passed |

The failures span every per-language unit-test population (`alterator::tests`, `checker::tests`, `ast::tests`, the per-metric modules) and the integration suites (`tests/grammars/`, `tests/output_formats/`, `tests/parity/`, the corpus tests `pdf_js_test` / `deepspeech_test`). None of these tests carries a `#[cfg(feature = "<lang>")]` gate; e.g. `alterator::tests::python_string_and_concatenated_string_flattened` runs under `--features rust` and panics on `LanguageDisabled(Python)`.

## Why CI cannot see it

The `feature-matrix` job in `.github/workflows/ci.yml` runs only

```
cargo check --all-targets ${{ matrix.flags }} --locked
```

on its non-default legs (`no-default-features (lib)`, `minimal-langs (lib)`, …). No CI leg *executes* a test under any partial feature set, so the "minimal grammar set" guarantee from #252 is compile-only, and the per-test gating discipline that `.claude/rules/testing.md` documents (the #1220/#1221 both-halves rule) is unenforceable — a regression that only manifests in a partial build can land silently, and the #1220 class recurs (nine sweeps already fail spuriously under the CI-named `rust,typescript` subset; see the companion issue).

## What is fine

[...]

Issue #1197, opened 2026-08-03
fix(npm): the #1184 function spaces emit npm/npa blocks their sibling methods do not


dekobon:
## Summary

The five constructs #1184 gave function spaces — Kotlin `get()` / `set()`
/ `init { … }`, Java and Groovy `static { … }`, JS `class_static_block` —
now emit `npm` and `npa` blocks that an ordinary method in the same class
does not.

```console
$ bca metrics -p C.kt
unit
  class    C        has: ['npm', 'npa']
    function <get>   has: ['npm', 'npa']    <-- all zeros, and new
    function m       has: []
```

## Cause

`Npm` and `Npa` self-enable on `Checker::is_func_space`:

```rust
if Self::is_func_space(node) && stats.is_disabled() {
    stats.is_class_space = true;
}
```

`is_func_space` is doing double duty. It means "this node opens a space",
which is what #1184 needed, but the OOP metrics read it as "this node is a
*container* that owns method/attribute counts". A method
(`MethodDeclaration`) is `is_func`, not `is_func_space`, so it never
tripped the second meaning — the five new constructs do.

## Impact

> **Wider than recorded here.** C#, JavaScript, MozJS, TypeScript, TSX,
> PHP and Ruby list ordinary methods in `is_func_space`, so every method
> in those languages already carried an all-zero block before #1184 —
> the JS-family `class_static_block` was one more instance, not a new
> inconsistency. Ten `Npm`/`Npa` impls used this shape and all ten
> moved.

Serialized-shape only: the added blocks are all zeros. But it is a
user-visible JSON shape change on every Kotlin, Java, Groovy and JS file
containing one of these constructs, and it is inconsistent — a `<get>`
space carries OOP metrics while the `m` beside it does not.

## Fix direction

> **Resolved.** The enable predicate wants to be "is this a scope that
> owns members", not "does this open a space". Note that *container* is
> the wrong word for it: `MetricScope::Container` excludes
> `SpaceKind::Unit`, so gating on it also deletes the whole-file
> roll-up (400 non-zero values across the integration corpus). The rule
> that fits is `wmc`'s — drop `Function` and `Unknown`, keep the rest —
> now shared as `SpaceKind::is_member_scope`.

The enable predicate wants to be "is this a *container* space", not "does
this open a space". The natural test is the space kind: `get_space_kind`
already returns `SpaceKind::Class` / `Interface` for real containers and
`Function` for these five.

[...]

Issue #1171, opened 2026-08-01
chore(dev): a fresh worktree needs two undocumented setup steps, one of which cannot be retried


dekobon:
## Summary

A fresh `git worktree` needs two setup steps that nothing states and nothing
detects, and one of them fails in a way that a plain retry cannot repair.

Both were hit repeatedly during the #1090-#1151 batch — five of twelve
worktree agents lost time to them, and two produced invalid measurements
before noticing.

## 1. The two missing setup steps

**`git submodule update --init --recursive`.** Without the integration corpus
under `tests/repositories/`, roughly 24-48 tests fail with errors that look
like real regressions in whatever the agent was changing.

**`make py-bootstrap`.** A fresh worktree inherits no `.venv`, so
`make pre-commit`'s `py-typecheck` stage reports ~33 mypy errors (`pytest`
untyped without its stubs) and `py-test` dies with "Couldn't find a
virtualenv". Both are bootstrap artifacts; all Python stages pass afterwards.

Neither failure names its cause, and both look like a broken change.

## 2. An interrupted submodule update is unrecoverable by repetition

This is the worse half. If `git submodule update --init --recursive` is
interrupted (it takes long enough to hit a command timeout), the submodule is
left with everything staged-deleted and an empty worktree — **and a plain
re-run is a silent no-op**, because the recorded SHA already matches HEAD.

The tell is `git status` showing the submodule at an unrelated SHA, or a
corpus directory containing only `snapshots/`. `--force` recovers it.

One agent's baseline timing run was invalid because of this and had to be
redone; another spent a debugging cycle on 5 "failing" corpus tests.

## Suggested direction

- [ ] A `make worktree-setup` (or extend an existing bootstrap target) that
      runs both steps idempotently, so there is one thing to run.
- [ ] Make the corpus-dependent tests fail with a diagnostic that names the
      cause — "integration corpus not checked out; run `git submodule update
      --init --recursive --force`" — rather than as N unrelated assertion
      failures. A single guard test asserting the corpus is populated would
      do it.
- [ ] Document both in `CLAUDE.md`'s worktree section and in the `batch-fix`
      skill's setup, including that recovery needs `--force`.

Related: #1145 fixed the third worktree papercut (`cargo fmt --all` could not
run there at all).

---

## Resolution

Fixed in `7b9a5ed4` on `fix/issue-1171`. All three checklist items are
done.

### 1. `make worktree-setup`

One idempotent target covering both steps: the integration corpora
[...]

PR #1479, opened 2026-09-16
test(features): gate every per-language test on the grammar it names

### LINKED ISSUE #1472: test(abc): the test module is ungated, so a minimal-feature build reports ~2,600 spurious failures
## Summary

`src/metrics/abc.rs`'s `mod tests` carries no feature gate, so **2,598 of
3,118** tests in it fail under `cargo test -p big-code-analysis
--no-default-features --features rust`. Every per-language test in the module
is compiled regardless of whether its grammar is enabled.

## Why it matters

`.claude/rules/testing.md` requires a `#[cfg(any(feature = ...))]` union gate
on any test whose fixture table is feature-dependent, so a build enabling
none of its rows drops the test rather than failing it. Individual modules
inside `abc.rs` now honour that — the newer ones added across 2026-09 carry
both the union gate and a non-vacuity guard. The **enclosing** module does
not, so the discipline is being applied test-by-test inside a container that
defeats it.

The practical effect is that a minimal-feature build cannot be used to verify
anything about ABC: the signal is buried under ~2,600 unrelated failures that
are bootstrap artifacts rather than regressions — the same class of
misleading-red as the missing-corpora case #1171 fixed.

## A second trap in the same area

A workspace-wide `cargo test --no-default-features --features X` does **not**
isolate features at all: `big-code-analysis-cli` and `big-code-analysis-web`
depend on the library with default features, and cargo unifies them, so
everything is silently re-enabled. Three different feature sets reported an
identical `4 passed; 3428 filtered out` before this was noticed — the
uniform-number tell from `.claude/rules/testing.md`.

Only `-p big-code-analysis` isolates. Any feature-gate verification run
without it proves nothing, and proves it while looking green. Worth a line in
`.claude/rules/testing.md` alongside the existing gate guidance.

## Suggested shape

Gate `mod tests` on the union of every language feature its contents name, or
split the per-language sections into already-gated child modules and gate the
parent on their union. Verify against a subset enabling none of them — note
`rust,typescript` is not such a subset for any table with a TypeScript or TSX
row, so a single non-listed language (`--features go`) is the reproducer.

## Related

- #1220 / PR #1221 — the same union-gate omission, twice in one batch.
- #1286 — the non-vacuity guard this pairs with.

---

## RESOLVED

All three items below are done on `fix/1472-gate-per-language-tests`
(12 commits, not pushed). Detail in the closing comment. In short:

1. **The 2,594 ungated tests** — re-measured as 3,025 across `src/`,
   `big-code-analysis-ast/src/` and `tests/`; 3,235 markers now, all but
   a handful derived.
2. **The gate that derives the marker** — `utils/check-test-lang-gates.py`,
   wired into `make pre-commit` / `make ci`. It reproduces all 167 gates
   written by hand in this tree, which is what licensed generating the
   rest.
3. **`Subject.carries_tests`** — fixed, with the matcher on the
   qualified module path.

The `feature-matrix` CI legs now run the suite rather than only
compiling it, which is what #1285 was waiting for.

One correction to the body below: the "30 of 3,482" figure does not
reproduce. 80 of 3,469 `src/` tests carried a feature `cfg` of their
own.

---

## Update — the CI half is closed; three items remain

PR #1476 closed the CI half of this issue. `.github/workflows/ci.yml`'s
`feature-matrix` job now documents that it runs clippy and not the suite,
and a new `utils/check-feature-gates.py` runs in each leg asserting that a
union-gated test subject is **absent** under a feature set enabling none of
its rows. What remains:

### 1. The 2,594 ungated per-language lib tests (the original issue)

Unchanged. Measured on `fix/batch-2026-09-14`: `--no-default-features
--features rust` fails **2,642 of 3,238** tests, 2,594 of them in the lib
target. Root cause is a contract violation, not a missing `cfg` in the
abstract:

```
panicked at big-code-analysis-ast/src/node.rs:84:
invariant: the caller checked LANG::is_enabled, or reached this through
AnyParser: LanguageDisabled(Python)
```

`ParserTrait::new`'s own doc says the same and directs callers to check
`LANG::is_enabled` first. Only **30 of 3,482** `#[test]` fns in `src/` do.
By module (halved for nextest's double-reporting): abc 525, loc 363,
cognitive 346, halstead 211, cyclomatic 207, npm 187, npa 180, wmc 167,
nargs 136, nexits 111, nom 77, tokens 36, mi 7.

Two approaches were considered and **rejected** during #1476; record them so
they are not re-proposed:

- **Gate each test.** Semantically right, ~2,600 attributes. The hazard is
  that a cross-language parity test needs `all(feature = …)` where a
  per-language test needs one feature, and a mechanical rewrite that gets
  that backwards produces a test that silently stops running.
- **Make the helpers self-skip on `is_enabled()`.** Only 3 of ~13 helpers
  take a runtime `LANG`; the rest take a type parameter, and `ParserTrait`
  exposes no language accessor, so it would need an additive change to a
  published trait. The disqualifier is independent of that: it would turn
  ~2,600 tests into silent no-ops under partial features, which
  `.claude/rules/testing.md` treats as worse than no test.

### 2. `check-feature-gates.py` cannot see a *missing* gate

A scanner of declared markers reports a clean tree when the marker is
deleted, because deleting it deletes the subject. That is the
`hidden_literal_supertypes` shape (fixed in `0e15c49c` on #1476), and only
its compile half — unused imports at `-D warnings` — was caught, by clippy.

The gate catches the **defeated** direction (a union present but weakened,
e.g. a non-feature disjunct), verified both ways during #1476. Catching the
**missing** direction needs a different gate: derive what the marker *should*
be from the union of the item's inner `#[cfg(feature = …)]` rows and compare
it against what is declared. That is the natural companion to item 1, since
the same derivation would drive a mechanical gating pass.

### 3. A latent false pass in `Subject.carries_tests`

`utils/check-feature-gates.py:265` classifies a `mod` as test-bearing only
when it carries its **own** bare `cfg(test)` predicate:

```python
return (self.kind == "mod" and self.is_cfg_test) or (
    self.kind == "fn" and self.is_test_fn
)
```

A union-gated `mod` nested inside an already-`#[cfg(test)]` parent has no
`test` in its own attribute, so it is classified `compile-time only` and
never compared against the nextest listing — a false pass in exactly the
#1220 shape the gate exists to catch. No current subject hits it (all 34 are
top-level `#[cfg(test)] mod`), so it is latent.

The naive repair is wrong: `subject_matcher` anchors on the bare name
(`(?:^|::)name(?:::|$)`), so admitting any `mod` makes a subject named
`tests` match every `::tests::` path in the crate and fail every disjoint
leg. Narrowing it needs the subject's full module path rather than its bare
name.

### Also noticed, adjacent

`src/test_support.rs:321` — `assert_csharp_fixture_spells` carries no
`#[cfg(feature = …)]` while its sibling `assert_perl_fixture_spells` (:329)
carries `#[cfg(feature = "perl")]`. Both are correct today: the C# callers
are ungated and the Perl ones are gated. It becomes `dead_code` at
`-D warnings` the moment a C# test module is gated the way
`perl_statement_modifier_parity` is — i.e. the first time item 1 is worked
on. Adding the gate before then would break the build.

dekobon:
Closes #1472, #1413, #1285, #1478.

Per-language tests named a concrete parser — `check_metrics::<PythonParser>(…)`
— without gating on that grammar. `mk_langs!` generates every `*Parser` alias,
`*Code` tag and `LANG` variant **unconditionally**, so such a test compiles with
`python` off and then panics in `Tree::new`. Measured on `main`: **3,025 tests
named a grammar without gating on it**, and `cargo nextest run
--no-default-features --features rust -p big-code-analysis` failed in the
thousands. A partial-feature build could not be used to verify anything — the
same misleading-red class as the missing-corpora case #1171 fixed.

## What landed

**`utils/check-test-lang-gates.py`** derives the marker an item needs from the
languages its body reaches, reading the language table out of the `mk_langs!`
invocation rather than keeping a copy. 3,500 gated items are checked on every
`make pre-commit`; `--fix` writes markers, `--show` prints the derivation.

**The `feature-matrix` CI legs now run the suite** instead of only compiling it
— every target but the three corpus-dependent binaries, whose fixtures live in
a submodule that job does not check out. That is the ratchet #1285 asked for,
and it is what makes the gate regress-testable.

**The gate checks both directions.** Too *narrow* is the silent one: too wide
panics on the leg that lacks the grammar, too narrow just drops the test and a
leg running two hundred fewer tests looks exactly like a green run. Sixteen
gates in the tree are deliberately wider than their bodies justify — a corpus
walk picking a language per file, a non-vacuity anchor, a `mod` declaration —
and each says so with a marker naming which features and why:

```rust
// test-lang-gates: hand-written(cpp) — the corpus walk picks a
//     language per file from its extension, so the glob list decides
//     it and nothing in the body names it
#[cfg(feature = "cpp")]
```

A gate that grows a feature nobody can account for fails, and so does a marker
that stops being load-bearing — otherwise the sixteen stop reading as a census.

**`--compare <ref>`** covers the residue. Everything above compares a marker
against the derivation; when the two agree and are *both* wrong, only the
previous revision says so — and that is computable statically. It scans the
tree at `ref` too (`git archive`, no cargo, no second build), derives which
single-language builds compile each test in each, and fails on a test that
still exists but stopped being built somewhere. The `lint` job runs it per PR
against the merge base. A *deliberate* narrowing looks identical and is waved
through with the **`gate-narrowing-intended`** label — not an in-source marker,
which would be permanently stale the moment this lands.

**`check-feature-gates.py`** (#1472 item 3) no longer misses a union-gated `mod`
nested inside an already-`#[cfg(test)]` parent — latent before this branch, and
about to stop being latent because the pass creates nested gated modules.

## Two derivation rules, both verified by revert

**A comparison is not a use.** `lang == LANG::Go` asks which variant a value is;
the enum is generated unconditionally, so it parses nothing. Reading it as a
requirement is what conjoined `feature = "go"` onto `containers_emit_npm_and_npa`
and dropped the positive half of the #1197 contract from every build without Go.
Deliberately *not* extended to the later arms of an or-pattern — `| LANG::X` is
also how a dispatch table groups its arms, and excluding those cost eleven items
their whole C-family union.

**A sweep still needs the parsers it hardcodes.** `is_enabled` filtering earns a
row set only `any(…)`; a parser named through a *type parameter* cannot be
skipped by any runtime filter, so it stays required even inside a sweep, and it
propagates through helpers exactly as `needs` does.

## Defects this found

- `tests/api/main.rs` gated `mod parser_reuse` on `all(rust, typescript)` when
  only three of its five tests need TypeScript. The other two had been dropped
  from every Rust-only build — before #1472, then blessed by a marker written
  during it. That leg now runs two tests it did not before.
- Six `Ccomment` / `Preproc` tests would have been gated out of every default
  build by an asymmetry between the root and ast crates' `c-family-helpers`
  features. Handled by `feature_atom`, which spells the atom per crate.

## Verification

- **Derivation validated against the 167 hand-written gates first**, before it
  was allowed to generate anything. If it could not reproduce what careful
  humans wrote over twelve months, the mechanical pass would have baked the
  error in 2,835 times.
- **`cargo nextest list --all-features --workspace` unchanged** — the set, not
  the count. The only difference against the baseline is ten `#[ignore]`d
  entries the two listings count differently; nothing was lost.
- **All 15 CI feature-matrix legs green locally**, 30/30 clippy and nextest.
- **70 self-tests** on the gate, including one per rule, per exclusion and per
  comparison spelling.
- `make pre-commit`: `BCA_GATE: pass`.

## Review notes

A fresh-context review of the over-gate check returned twelve findings, four of
which changed a verdict and three of which were invisible from the gate's own
output — it reported clean either way. The over-gating rule collapsed from a
rule-plus-two-exclusions into one; `over_gated` learned to skip `pub` items;
`hardcoded` learned to propagate through helpers; and `--fix` learned to write a
marker that satisfies the check that asked for it (it previously stacked four
`#[cfg]`s on a sweep-that-pins and exited 2). Five of the nineteen markers the
first cut accepted were artefacts of those defects — hence sixteen.

The last two commits are unrelated to the gate: a `.claude/rules/shell.md` entry
for a `pgrep -f` wait loop that matches its own command line and never exits,
which leaked ten stuck waiters during this work, plus the review-driven
correction to it.

## Not covered

The diff is dominated by ~3,235 mechanical `#[cfg]` attribute lines across 65
`.rs` files. Worth a `.git-blame-ignore-revs` entry if this repo starts keeping
one.


dekobon:
## `--compare` failed, found a real bug, and then needed the label

Recording this rather than just applying the label, because a silently
waved-through check is the hole this gate exists to prevent.

### It caught something

`--compare` ran in CI for the first time on this PR and reported two
`every_*_in_every_language` parity sweeps gated
`any(c, c-family-helpers, cpp, mozcpp)`. Both walk the whole `LANG`
enum. They were running in four builds instead of twenty-three — and
had been since before this branch.

Neither other check could see it. `over_gated` compares a gate against
the derivation, so a gate faithfully mirroring a *wrong* derivation is
by construction never flagged, and an `any(...)` requires no single
feature for it to object to. That is the exact argument for `--compare`
existing, made on its first run.

Two derivation defects, both now fixed at the root:

- **Every alternative of a `matches!` is an identity test**, not just
  the leading one. `matches!(lang, LANG::Ccomment | LANG::Preproc)`,
  written to *skip* two languages, read as a requirement for one of
  them — and that single leak was the entire gate on both sweeps. This
  is not the or-pattern caveat, which is about `match` *arms* where `|`
  groups a dispatch table a sweep legitimately derives from.
- **A full-enum sweep needs `any(<every language>)`**, when it carries
  both halves of the `.claude/rules/testing.md` rule (an `is_enabled()`
  filter *and* a non-vacuity assertion) — those together mean it fails
  rather than skips with no language enabled. Conditioned on that,
  because `Display` / `FromStr` / slug sweeps walk the same enum, parse
  nothing, and belong on the `--no-default-features` leg.

### Then it still failed, and the label is correct

With those fixed, `--compare` reports **3,001** tests that lost a build
since the merge base. That is the point of the PR: those tests were
ungated, so every probe build compiled them — and then they *panicked*
in `Tree::new`. "Stopped being built by `--features go`" means "stopped
panicking under `--features go`".

I did not take that on faith. Each of the 3,001 was checked against its
derived needs, asking whether the lost build could actually have run it:

```
  legitimate (lost only builds it could not have run in): 3001
  SUSPICIOUS (lost a build it could have run in)         : 0
  no derived needs / not found                          : 0
```

Independent corroboration: all 15 feature-matrix legs now **run** the
suite green where they previously only compiled it, and
`cargo nextest list --all-features` is byte-identical to `main`, so
nothing left the full build.

This is the bootstrap case — the PR that introduces the gate is the one
PR whose whole purpose is wholesale narrowing — so
`gate-narrowing-intended` is applied. Worth noting the label did not
exist until this PR needed it; the escape hatch was documented and
wired into the CI condition but never created, so the first legitimate
narrowing would have hit a blocking check whose documented remedy
silently never matched.

### Fixed in `3557fd8c`

- 15 legs green, 30/30 clippy and nextest
- `nextest list --all-features` unchanged
- 73 self-tests, including one per new rule and the negative case
- `make pre-commit`: `BCA_GATE: pass`

dekobon:
## Independent review (fable): 13 findings, 12 fixed in `ed26071b`, 1 filed as #1480

### The blocker was real

The `assert_eq!` comparison rule matched with `[^()]*?`, which cannot
cross a parenthesis, so it stopped at the first call in the argument
list and never reached the variant:

```rust
assert_eq!(guess_language(buf, "run"), (Some(LANG::Ruby), "ruby"));
```

`guess_language` / `get_from_ext` / `FromStr` / `name()` are not
`cfg`-gated — they read the table `mk_langs!` generates unconditionally
— so **36 tests that parse nothing were gated on grammars they never
touch**. `test_guess_language` ran in one build of twenty-five.

Nothing on this branch could see it, and that is the uncomfortable part:
`over_gated` compares a gate against the derivation, so a marker
mirroring a wrong derivation is never flagged; and `--compare`, the one
check that would have caught it, is skipped here by the
`gate-narrowing-intended` label. **The escape hatch hid a real
regression on its first use.**

Worse, the triage I posted earlier in this PR — "3,001 lost builds, 0
suspicious" — was computed from the same derivation, so it was blind in
exactly the same way and reported these 36 as legitimate. With the fix
in, the same triage reads **2,967 lost, 0 suspicious**, and the 34-test
difference is those tests returning to every build.

### Also fixed, each with a regression test

- A variant in receiver position reads metadata (`LANG::Mozcpp
  .extensions()`), **except** `tree_sitter_language()` / `get_ts_language()`,
  which return `None` with the feature off and are `.expect()`ed — the
  `Tree::new` panic one call earlier.
- `all(not(a), not(b))` made every feature read as over-declared:
  probing one at a time never finds the build such a gate admits.
- `#![cfg(...)]` was invisible — all three readers tested `#[` first.
- `cfg(not(test))` counted as a test scope.
- `=> Some(LANG::X)` was not a dispatch arm, so a `-> Option<LANG>`
  helper handed every language to each caller.
- `--compare`'s reference scan had no non-vacuity guard: an empty
  `before` printed "OK — no test lost a build".
- `UnicodeDecodeError` escaped the handler and exited 1, which reads as
  "offenders found".

Docs: the quoted 3,025 ungated tests re-measures to ~2,950 and is now
marked approximate (it moves with the derivation rules themselves);
three claims about deriving gates for *imports*, which the tool has
never done, are gone; "merge base" is now "base branch head", which is
what `base.sha` is.

### Two gates I removed and had to put back

Both were load-bearing, and one was caught only by the
`--no-default-features` leg:

- `from_path_tests` picks its grammar from a `"foo.rs"` extension
  string — the documented blind spot.
- `ast_from_tree_sitter_returns_language_disabled_for_off_feature`
  builds a real Rust tree while requiring JavaScript off.

Both now carry a `hand-written` marker. I also briefly wrote a marker
claiming one of them used `tree_sitter_rust` directly; it does not, and
that comment is gone along with the over-broad receiver rule that
prompted it.

### Left open

#1480 — a sweep's exemption also covers a `LANG` it names outside the
loop. Pinning every literal was tried and **reverted**: it breaks 22
items and 8 self-tests, because a sweep's row table is itself written as
literals and those *are* skipped by the runtime filter. All 16 live
sweeps filter correctly; the boundary is documented in
`.claude/rules/testing.md`.

### Verification

81 self-tests · 15 legs green (30/30 clippy and nextest) ·
`nextest list --all-features` byte-identical · `BCA_GATE: pass`

